PR #2, opened 2015-04-06
Validation and conversion support

lvonk:
Validation and conversion was a bit messy. This is the first attempt to clean up and avoid usual repetition when creating Commands and ValueObjects.

Prior to this PR you have to:
```
class MyCommand < Sequent::Core::BaseCommand
    attrs number: Integer, 
          date: Date, 
          date_time: DateTime, 
          list_of_numbers: array(Integer), 
          user: Person
    
    validates_numericality_of :number, only_integer: true
    #custom validator for date needed
    #custom validator for datetime needed
    validates_with Sequent::Core::AssociationValidator, associations: [:list_of_numbers, :user]

end
```

And parse Integers and DateTime yourself. This made form binding a bit cumbersome and hard to work with.

With this PR all expected validations - like Integers being an Integer (or valid string as Integers), Dates being a Date (or valid string as Dates) etc etc - are added by default. Presence is still optional and you should still add it yourself if needed.

So the above can be rewritten as follows:
```
class MyCommand < Sequent::Core::BaseCommand
    attrs number: Integer, 
          date: Date, 
          date_time: DateTime, 
          list_of_numbers: array(Integer), 
          user: Person
end
```

Furthermore the `CommandService` will parse all values to their correct types (the supported ones are in the README) after the `Command` is `valid?`.


lib/sequent/core/helpers/type_conversion_support.rb
```diff
@@ -0,0 +1,27 @@
...
+
+module Sequent
+  module Core
+    module Helpers
+      # Will parse all values to the correct types.
+      # The raw values are typically posted from the web and are therefor mostly strings.
+      # To parse a raw value your class must have a parse_from_string method that returns the parsed values.
+      # See sequent/core/ext for currently supported classes.
+      module TypeConversionSupport
+        def parse_attrs_to_correct_types!
```
bforma: I expect a `Command` to be immutable. Thus, I would prefer to rename `parse_attrs_to_correct_types!` to `parse_attrs_to_correct_types` and let it return a copy of the command with it's values (read: types) changed.

lib/sequent/core/ext/ext.rb
```diff
@@ -0,0 +1,118 @@
+class Symbol
```
derekkraan: Is it possible to avoid monkey-patching core classes? It's not exactly gem best-practice.

derekkraan: ... I see that you've just moved this code, but the point still stands ;)

lib/sequent/core/helpers/type_conversion_support.rb
lvonk: Yes that would be better

lvonk:
Moved monkey patches to normal classes and made command immutable. There is still one monkey patch left that was put in for performance reason for parsing from json. We need to verify if this is still a valid argument.

lib/sequent/core/helpers/association_validator.rb
```diff
@@ -26,7 +26,16 @@ def validate(record)
             if value && !value.kind_of?(Array) && record.respond_to?(:attributes) && !value.kind_of?(record.attributes[association])
               record.errors[association] = "is not of type #{record.attributes[association]}"
             elsif value && value.kind_of?(Array)
-              record.errors[association] = "is invalid" if value.any? { |v| not v.valid? }
+              item_type = record.class.type_for(association).item_type
+              record.errors[association] = "is invalid" if value.any? do |v|
```
derekkraan: this inline `if` seems less handy now that it's got a multiline block on the end of it

lib/sequent/core/helpers/value_validators.rb
```diff
@@ -0,0 +1,59 @@
...
+        VALIDATORS = {
+          ::Symbol => ->(_) { true },
+          ::String => ->(value) { value.nil? || value.is_a?(String) },
+          ::Integer => ->(value) { valid_integer?(value) },
+          ::Boolean => ->(value) { valid_bool?(value) },
+          ::Date => ->(value) { valid_date?(value) },
+          ::DateTime => ->(value) { valid_date_time?(value) }
+        }
+
+        def self.valid_integer?(value)
```
derekkraan: this could be written:
```def self.valid_integer?(value)
  return true if value.blank?
  Integer(value)
  true
rescue
  false
end
```

lib/sequent/core/helpers/string_to_value_parsers.rb
```diff
@@ -0,0 +1,76 @@
...
+          ::String => ->(value) { value },
+          ::Integer => ->(value) { parse_to_integer(value) },
+          ::Boolean => ->(value) { parse_to_bool(value) },
+          ::Date => ->(value) { parse_to_date(value) },
+          ::DateTime => ->(value) { parse_to_date_time(value) },
+          ::Sequent::Core::Helpers::ArrayWithType => ->(values, type_in_array) { parse_array(values, type_in_array) }
+        }
+
+        def self.parse_to_integer(value)
+          Integer(value) unless value.blank?
```
derekkraan: We never do anything with `Integer(value)`. Is it necessary?

lib/sequent/core/helpers/param_support.rb
```diff
@@ -49,8 +47,10 @@ def as_params
               value = value.as_params
             elsif value.kind_of?(Array)
```
derekkraan:
```
case value
when Array
  ...
when DateTime
  ...
when Date
  ...
end
```
maybe useful here?

lib/sequent/core/command_service.rb
```diff
@@ -72,13 +72,13 @@ def execute_commands(*commands)
...
 
               if command.valid?
-                @command_handlers.each do |command_handler|
-                  command_handler.handle_message command if command_handler.handles_message? command
-                end
+                parsed_command = command.parse_attrs_to_correct_types
+                @command_handlers.select { |h| h.handles_message?(parsed_command) }.each { |h| h.handle_message parsed_command }
+                @repository.commit(parsed_command)
+              else
+                raise CommandNotValid.new(command)
```
derekkraan: you could also do early return and avoid an if/else:

`raise CommandNotValid.new(command) unless command.valid?` before the current if/else.

lib/sequent/core/helpers/date_time_validator.rb
```diff
@@ -0,0 +1,22 @@
...
+module Sequent
+  module Core
+    module Helpers
+      # Validates DateTimes
+      # Automatically included when using a
+      #
+      #   attrs value: DateTime
+      class DateTimeValidator < ActiveModel::EachValidator
+        def validate_each(subject, attribute, value)
+          return if value.is_a?(DateTime)
```
derekkraan: I'd also consider doing it like this:

```ruby
def validate_each(subject, attribute, value)
  return if value.is_a?(DateTime)
  DateTime.deserialize_from_json(value)
rescue
  subject.errors.add attribute, :invalid_date_time
end
```

lib/sequent/core/helpers/date_time_validator.rb
```diff
@@ -0,0 +1,22 @@
...
+          begin
+            DateTime.deserialize_from_json(value)
+          rescue
```
derekkraan: I'm not 100% sure why we use rescue here. Some of our `deserialize_from_json` values do throw raises, and some do not. It depends on the implementation of the underlying objects. Integer, for example, tends to return 0 if it finds something weird:

`"".to_i
=> 0
`

It seems better to catch these type of things in `deserialize_from_json` and there to throw a better error, which we can then catch in this method.

My second comment is that `rescue` (implicitly all exceptions) is always a code smell for me.

derekkraan:
This is also relevant: https://codeclimate.com/github/zilverline/sequent/compare/validation-conversion#ratings

lib/sequent/core/helpers/date_time_validator.rb
lvonk: I'd like to look at the deserialisers separately. The `"".to_i => 0` is indeed not want we want.

For the rescue: I think in a validator it is okay. For what ever reason we can't parse into a DateTime.

lib/sequent/core/helpers/association_validator.rb
lvonk: Refactored

lib/sequent/core/helpers/value_validators.rb
lvonk: :+1:

lib/sequent/core/helpers/string_to_value_parsers.rb
lvonk: Nope. Deleted the deserialize since that results in `"".to_i`  being 0

lib/sequent/core/helpers/param_support.rb
lvonk: Probably. I'd like to look at `ParamSupport` separately.

lib/sequent/core/helpers/date_time_validator.rb
lvonk: :+1:

README.md
```diff
@@ -66,6 +66,18 @@ Commands in sequent use ActiveModel for validations.
...
+* Date
+* DateTime
+* Integer
+* Boolean
+
+When posted form the web and created like `PayInvoiceCommand.from_params(params[:pay_invoice_command])` its values are typically still all String.
+Sequent takes care of parsing the string values to the correct types automatically when a Command is valid (`valid?`).
+Its values will be parsed to the correct types by the CommandService. If you instantiate a Command manually with the correct types nothing will change.
+
+
```
bforma: Surplus newline

lib/sequent/core/command_service.rb
```diff
@@ -71,14 +71,11 @@ def execute_commands(*commands)
...
-                end
-              end
-
-              @repository.commit(command)
-              raise CommandNotValid.new(command) unless command.validation_errors.empty?
+              raise CommandNotValid.new(command) unless command.valid?
+              parsed_command = command.parse_attrs_to_correct_types
+              @command_handlers.select { |h| h.handles_message?(parsed_command) }.each { |h| h.handle_message parsed_command }
+              @repository.commit(parsed_command)
+
```
bforma: Surplus newline

lib/sequent/core/helpers/association_validator.rb
```diff
@@ -23,13 +23,29 @@ def validate(record)
...
+              item_type = record.class.type_for(association).item_type
+              record.errors[association] = "is invalid" if all_valid?(value, item_type)
             else
               record.errors[association] = "is invalid" if value && value.invalid?
+            end
+          end
+        end
 
+        private
+        def incorrect_type?(value, record, association)
```
bforma: Could use a newline

lib/sequent/core/helpers/association_validator.rb
```diff
@@ -23,13 +23,29 @@ def validate(record)
...
+          !value.kind_of?(Array) && record.respond_to?(:attributes) && !value.kind_of?(record.attributes[association])
+        end
+        def all_valid?(value, item_type)
```
bforma: Could use a newline

lib/sequent/core/helpers/string_to_value_parsers.rb
```diff
@@ -0,0 +1,75 @@
...
+        def parse_from_string(value)
+          parser = PARSERS[@klass]
+          if @array_with_type
+            parser.call(value, @array_with_type.item_type)
+          else
+            parser.call(value)
+          end
+        end
+      end
+
```
bforma: Surplus newline

lib/sequent/core/helpers/string_to_value_parsers.rb
```diff
@@ -0,0 +1,75 @@
...
+    end
+  end
+end
+
```
bforma: Surplus newline

lib/sequent/core/helpers/value_validators.rb
```diff
@@ -0,0 +1,55 @@
...
+
+        def initialize(klass)
+          @klass = klass
+        end
+
+        def valid_value?(value)
+          VALIDATORS[@klass].call(value)
+        end
+      end
+
```
bforma: Surplus newline

lib/sequent/core/helpers/value_validators.rb
```diff
@@ -0,0 +1,55 @@
...
+    end
+  end
+end
+
```
bforma: Surplus newline

spec/core/command_service_spec.rb
```diff
@@ -49,4 +49,29 @@ class DummyBaseCommand < Sequent::Core::BaseCommand
...
+    end
+
+    it "does not parse values if the command is invalid" do
+      command = WithIntegerCommand.new(value: "A")
+      expect { command_service.execute_commands(command) }.to raise_error do |e|
+                                                                expect(e.errors[:value]).to eq ['is not a number']
+                                                              end
+    end
+  end
+
```
bforma: Surplus newlines

derekkraan:
:+1:

PR #3, opened 2015-04-10
Automatically register event / command handlers

derekkraan:
Problem: event and command handlers are order-dependent in Freemle. (D'oh!)

### DIFF
--- a/lib/sequent/core/command_service.rb
+++ b/lib/sequent/core/command_service.rb
@@ -2,20 +2,6 @@ require_relative 'transactions/no_transactions'
 
 module Sequent
   module Core
-
-    class CommandServiceConfiguration
-      attr_accessor :event_store,
-                    :command_handler_classes,
-                    :transaction_provider,
-                    :filters
-
-      def initialize
-        @command_handler_classes = []
-        @transaction_provider = Sequent::Core::Transactions::NoTransactions.new
-        @filters = []
-      end
-    end
-
     #
     # Single point in the application where subclasses of Sequent::Core::BaseCommand
     # are executed. This will initiate the entire flow of:
@@ -27,22 +13,6 @@ module Sequent
     # * Unit of Work is cleared
     #
     class CommandService
-      class << self
-        def configure
-          configuration = instance.configuration.dup
-          yield(configuration) if block_given?
-          @instance = new(configuration) # this makes it threadsafe
-        end
-
-        def reset
-          @instance = new
-        end
-
-        def instance
-          @instance ||= new
-        end
-      end
-
       attr_accessor :configuration
 
       # +DefaultCommandServiceConfiguration+ Configuration class for the CommandService containing:
@@ -53,11 +23,6 @@ module Sequent
       #   +filters+ List of filter that respond_to :execute(command). Can be useful to do extra checks (security and such).
       def initialize(configuration = CommandServiceConfiguration.new)
         self.configuration = configuration
-        @event_store = configuration.event_store
-        @repository = AggregateRepository.new(configuration.event_store)
-        @filters = configuration.filters
-        @transaction_provider = configuration.transaction_provider
-        @command_handlers = configuration.command_handler_classes.map { |handler| handler.new(@repository) }
       end
 
       # Executes the given commands in a single transactional block as implemented by the +transaction_provider+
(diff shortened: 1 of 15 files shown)


lvonk:
I can see this working for CommandHandlers, but some event handlers have collaborators. For now I prefer registering them explicit. Or at least have the option to choose for automagic or not.

erikrozendaal:
I definitely prefer explicit registration of command and event handlers. Also the mechanism implemented in this pull request also mutates a global variable and cannot be overridden or replaced.

If automatic registration is wanted maybe add a method to the `CommandService` and the `EventStore` to automatically find and register all handlers using something like http://stackoverflow.com/a/2393750?

Even better, extract dispatching commands and events this into a separate class (`MessageBus`?) to remove duplication.

derekkraan:
@erikrozendaal just leaving your other points unanswered for a moment, if you check the implementation of `CommandService.configure` (and `EventStore.configure`), it should be threadsafe (we mutate a duped configuration, and then replace the entire instance with a newly configured instance).

derekkraan:
To answer another of your points:  extracting the functionality looks like a good idea to me.

derekkraan:
Your last point (with SO link): then you require that all event listeners and command handlers are subclasses of the base class. Maybe that's what you want, but maybe not?

erikrozendaal:
If I understand the code correctly by just reading it, a new configuration instance is mutated each time the base class is inherited. Does automatic registration work at all when multiple handlers are defined?

My preference is still manual registration only. But I can live with optional automatic registration.

Maybe by adding a method to `CommandService` and `EventStore` (configuration?) like `automatically_register_handlers_of_kind(class=SomeHandyDefault)`.

I'm not sure I understand your point on having to extend a specific base class. Isn't the current solution similar?

derekkraan:
```ruby
def configure
  configuration = instance.configuration.dup
  yield(configuration) if block_given?
  @instance = new(configuration) # this makes it threadsafe
end
```

So if you do `configure do |config| config.foo << :bar end`, you are mutating a `dup`ed configuration, then creating a new instance with that configuration, and then replacing (not mutating!) `@instance`.

There are tests to make sure that you can register multiple handlers; according to those tests (and my understanding of the code), it works.

erikrozendaal:
Ah, I looked at the master branch code...

derekkraan:
:+1:

Maybe the automatic registration should work like this:

* automatic registration
* possible to register manually + disable further automatic registration in an initializer.

I also wanted to talk to @lvonk about the things in Freemle that are standing in the way of automatic registration and how (/ if) we can fix that.

erikrozendaal:
Isn't `dup` shallow? In that case you're still mutating the shared `command_handler_classes` list in https://github.com/zilverline/sequent/blob/auto_register_command_handlers/lib/sequent/core/base_command_handler.rb#L42.

lib/sequent/core/command_service.rb
```diff
@@ -28,22 +27,24 @@ def initialize
...
     class CommandService
-
       class << self
-        attr_accessor :configuration,
-                      :instance
-      end
+        def configure
+          configuration = instance.configuration.dup
+          yield(configuration) if block_given?
+          @instance = new(configuration) # this makes it threadsafe
```
erikrozendaal: Does Ruby (the language, not the MRI) guarantee thread-safety of instance variable assignment? Also, when multiple threads are configuring concurrently some configuration will be lost (last thread wins?).

Actually, I don't think thread-safety is a real concern here (all handlers are probably registered at load time in a single thread), but the comment might be misleading.

derekkraan:
Yes you're right. I'll have to fix that.

lvonk:
Maybe we should split this in 2 things:

1. Refactor configuring sequent.
2. Implement optional auto-discovery of event handlers and command handlers 

**1. Refactor configuring sequent**
Something like:

```
Sequent.configure do |config|
  config.command_handlers = [...]
  config.event_handlers = [...]
end

# Then access like this
Sequent.command_service
``` 

**2. Implement optional auto-discovery of event handlers and command handlers **

Like Derek suggest and in the new config style:

```
Sequent.configure do |config|
  config.autodiscover_event_handlers = true
  config.autodiscover_command_handlers = true
  config.autodiscover_all = true
end
```

derekkraan:
@lvonk and @erikrozendaal, what do you think? It works basically the way that @lvonk suggested. Auto-discovery is on by default, though.

One other thing, I'm still debating whether to use `.descendants` or let the class register itself. On the one hand, I like the idea of leaving that responsibility to the classes, but on the other hand, if you do `Sequent::Configuration.new` then you won't have the auto-discovered classes anymore.

lvonk:
I would like to disable by default (at least for now), plus could you add documentation in the README on manual configuration?

erikrozendaal:
Both auto-discovery methods seems to present some load ordering problems. Using `.descendents` you have to ensure all handlers are loaded before you configure sequent. Using `self.inherited` you cannot easily reset the configuration.

Hot reloading in development will probably cause even more trouble?

lvonk:
Good points @erikrozendaal. @derekkraan I'd really like the auto discovery as a separate PR from refactoring the Sequent config. I am also not really convinced yet that auto-discovery of command handlers and event handlers is a good idea at this point.

derekkraan:
Closing ... might resurrect this PR later.

PR #5, opened 2015-04-19
Performance improvement when using oj instead of to_json

lvonk:
Test serializing > 500.000 events to_json:

With to_json: [68921] ms
With oj: [21171 ms]

Removing ActiveSupport::JSON completely as a result calling `valid?` on a `Command` or `ValueObject` will not result in serialized `errors` and `validation_context` attributes.

```
person = Person.new(name: "Kim")
person.valid?
Sequent::Core::Oj.dump(person)
=> "{\"name\":null,\"validation_context\":null,\"errors\":{}}"
Sequent::Core::Oj.dump(person.attributes)
=> "{\"name\":null}"
```

Both `EventRecord` and `CommandRecord` will call `attributes` method on `event` and `command` before persisting.

### DIFF
--- a/lib/sequent/core/event.rb
+++ b/lib/sequent/core/event.rb
@@ -29,7 +29,7 @@ module Sequent
       end
       protected
       def payload_variables
-        %w{@aggregate_id @sequence_number @created_at @underscored}
+        %w{@aggregate_id @sequence_number @created_at}
       end
 
     end
--- a/lib/sequent/core/event.rb
+++ b/lib/sequent/core/event.rb
@@ -22,7 +22,10 @@ module Sequent
 
       def payload
         result = {}
-        instance_variables.reject { |k| payload_variables.include?(k.to_s) }.each do |k|
+        instance_variables
+          .reject { |k| payload_variables.include?(k.to_s) }
+          .select { |k| attributes.keys.include?(to_attribute_name(k))}
+          .each do |k|
           result[k.to_s[1 .. -1].to_sym] = instance_variable_get(k)
         end
         result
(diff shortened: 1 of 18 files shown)


lib/sequent/core/event.rb
```diff
@@ -24,14 +22,22 @@ def initialize(args = {})
 
       def payload
         result = {}
-        instance_variables.reject { |k| payload_variables.include?(k.to_s) }.each do |k|
+        instance_variables
+          .reject { |k| payload_variables.include?(k.to_s) }
```
erikrozendaal: Maybe `payload_variables` should be an array of symbols (`%i{...}`) instead of strings? Saves a `to_s` call here.

lib/sequent/core/event.rb
```diff
@@ -24,14 +22,22 @@ def initialize(args = {})
...
       end
       protected
       def payload_variables
-        %w{@aggregate_id @sequence_number @created_at @underscored}
+        %w{@aggregate_id @sequence_number @created_at}
+      end
+
+      private
+      def to_attribute_name(instance_variable_name)
+        instance_variable_name.to_s.gsub('@', '').to_sym
```
erikrozendaal: Only remove the leading `@` instead of replacing all with the empty string. The slice operator (`s[1..-1]`) also works on symbols, so the the `to_s` can be removed (you still need to `to_sym` though).

lib/sequent/sequent.rb
```diff
@@ -1,3 +1,7 @@
+require 'oj'
+Oj.default_options={mode: :compat}
```
erikrozendaal: Strange place for this configuration option? Can't we instantiate a global `sequent_oj` and use that everywhere?

lvonk:
@erikrozendaal better now?

erikrozendaal:
:+1:

PR #6, opened 2015-04-24
Improve configuration

derekkraan:


README.md
```diff
@@ -20,13 +20,9 @@ If you are unfamiliar with these concepts you can catch up with:
...
 
-    Sequent::Core::EventStore.configure do |config|
-      config.event_handlers = [MyEventHandler.new]
-    end
-
-    Sequent::Core::CommandService.configure do |config|
-      config.event_store = Sequent::Core::EventStore.instance
-      config.command_handler_classes = [MyCommandHandler]
+    Sequent.configure do |config|
+      config.event_handlers = [MyEventHandler]
```
lvonk: Shouldn't this be `[MyEventHandler.new]` ?

lvonk:
We should probably bump version to `0.2.0` since it is not backward compatible?

README.md
derekkraan: Nope. The `.new` happens elsewhere now.

derekkraan:
IMO that's a decision you would make when deciding to release. Sounds logical though.

lib/sequent/core/event_store.rb
```diff
@@ -72,15 +48,20 @@ def replay_events
...
       protected
+
       def record_class
-        @record_class
+        configuration.record_class
+      end
+
+      def event_handlers
+        configuration.event_handlers
       end
```
lvonk: @derekkraan shouldn't this than be `event_handlers.map(&:new)`?

lib/sequent/core/event_store.rb
```diff
@@ -72,15 +48,20 @@ def replay_events
...
 
       protected
+
       def record_class
-        @record_class
+        configuration.record_class
+      end
+
+      def event_handlers
+        configuration.event_handlers
```
lvonk: @derekkraan shouldn't this than be `event_handlers.map(&:new)`?

README.md
lvonk: I can't find that anywhere in the code... See not below. Also in our current implementation we have some `EventHandler`s with collaborators. How would that work in this situation.
I do agree that both `CommandHandler`s and `EventHandler`s should be registered the same way. So either both as instances (my preferences) or both both only the classes. What do you think?

derekkraan:
@lvonk I'm convinced. Any other feedback for this PR?

lvonk:
Nope. Let's merge!

PR #9, opened 2015-06-02
Added an new spec helper method to assert errors 

acarpe:
...and restored given_events with new syntax, I've updated the relative examples


erikrozendaal:
Thanks for your pull request. We've made a separate fix for the `given_events` issue (see https://github.com/zilverline/sequent/pull/10).

lib/sequent/test/command_handler_helpers.rb
```diff
@@ -26,11 +26,16 @@ module Test
...
-    #     given_events InvoiceCreatedEvent.new(args)
+    #     given_events Invoice, invoice_id, InvoiceCreatedEvent.new(args)
     #     when_command PayInvoiceCommand(args)
     #     then_events InvoicePaidEvent(args)
     #   end
     #
+    #   it "rejects coupon when does not exist" do
+    #     given_events Invoice, invoice_id, CartCreatedEvent.new(args)
+    #     when_command ApplyDiscountCouponCommand(args)
+    #     then_fail_with CouponDoesNotExist
```
bjpbakker: `then_fails_with`

lib/sequent/test/command_handler_helpers.rb
```diff
@@ -65,7 +70,7 @@ def commit_events(_, streams_with_events)
           end
         end
 
-        def given_events(streams_with_events)
+        def given_events(*streams_with_events)
```
bjpbakker: A fix for this is made in master. We like using just a sequence of events rather than event streams in specs.

lib/sequent/test/command_handler_helpers.rb
```diff
@@ -84,17 +89,26 @@ def deserialize_events(events)
...
       end
 
+
       def when_command command
         raise "@command_handler is mandatory when using the #{self.class}" unless @command_handler
         raise "Command handler #{@command_handler} cannot handle command #{command}, please configure the command type (forgot an include in the command class?)" unless @command_handler.handles_message?(command)
-        @command_handler.handle_message(command)
+        begin
+          @command_handler.handle_message(command)
+        rescue Sequent::Core::DomainError => e
```
bjpbakker: Why scope this to `Sequence::Core::DomainError`? IMO we should be able to match all exceptions.

lib/sequent/test/command_handler_helpers.rb
```diff
@@ -84,17 +89,26 @@ def deserialize_events(events)
...
+          @actual_error = e
+        end
         @repository.commit(command)
       end
 
+      def then_fails_with clazz
+        expect(@actual_error.class).to be(clazz)
+      end
+
       def then_events *events
```
bjpbakker: When `@actual_error` is present, raise it.

lib/sequent/test/command_handler_helpers.rb
```diff
@@ -84,17 +89,26 @@ def deserialize_events(events)
...
+        begin
+          @command_handler.handle_message(command)
+        rescue Sequent::Core::DomainError => e
+          @actual_error = e
+        end
         @repository.commit(command)
       end
 
+      def then_fails_with clazz
+        expect(@actual_error.class).to be(clazz)
```
bjpbakker: `expect(@actual_error).to be_kind_of(class)` ?

bjpbakker:
Thanks for your pull request. Expecting errors in the same way we expect events would make these specs more readable. I've added some comments.

Also please merge master into your PR.

lib/sequent/test/command_handler_helpers.rb
acarpe: I'm scoping to Sequence::Core::DomainError in order to make other kind of errors to flows throw the test and interrupt the test execution.
But yes I can make it without scoping, I'll change it.

lib/sequent/test/command_handler_helpers.rb
acarpe: If I raise the @actual_error it could cause the failing of the test altough the error was the expected right one

bjpbakker: I would expect a command to either fail or result in new events, but not both. An error would result in a transaction rollback anyway.

As errors are now silently ignored on executing the command, it is helpful to raise them (with their original stack) as `then_events` is called. Otherwise tests may pass where the code actually raises an error.

acarpe: Ah ok, now I got it.

lib/sequent/test/command_handler_helpers.rb
```diff
@@ -114,14 +119,24 @@ def given_events *events
...
       end
 
+
       def when_command command
         raise "@command_handler is mandatory when using the #{self.class}" unless @command_handler
         raise "Command handler #{@command_handler} cannot handle command #{command}, please configure the command type (forgot an include in the command class?)" unless @command_handler.handles_message?(command)
-        @command_handler.handle_message(command)
+        begin
+          @command_handler.handle_message(command)
+        rescue StandardError => e
```
bjpbakker: `rescue => e` only rescues `StandardError` derived errors; no need to specify that here.

lib/sequent/test/command_handler_helpers.rb
```diff
@@ -114,14 +119,24 @@ def given_events *events
...
+        end
         @repository.commit(command)
       end
 
+      def then_fails_with clazz
+        expect(@actual_error).to be_kind_of(clazz)
+      end
+
       def then_events *events
+        raise @actual_error unless !@actual_error
```
bjpbakker: `raise @actual_error if @actual_error`

lvonk:
@bjpbakker what is the status of this PR? Can we merge it?

bjpbakker:
It's fine now. Thanks for the improvements @acarpe, and for this PR in the first place

PR #10, opened 2015-06-05
Restore old `given_events` API

erikrozendaal:
Automatically determine the associated event stream information by
looking at the event handlers defined in the aggregate roots. This
allows the old test syntax to be used when defining test scenarios.


bjpbakker:
:+1:

PR #12, opened 2015-06-23
Give `.from_params` an overrideable constructor

derekkraan:
`.from_params` is already a constructor, just like `.new`, so give it
its own overrideable `#initialize`, simply called `#from_params`.


lvonk:
:+1:

PR #13, opened 2015-07-08
Preserve hash implementation when merging

bjpbakker:
If a `HashWithIndifferentAccess` is passed, merging it into a regular hash
will change semantics. Now `merge` is called on the passed-in hash.

Also, it's a bit safer to eliminate possible overrides of `aggregate_id`
and `sequence_number`.


bforma:
:+1:

PR #14, opened 2015-07-08
Default `handles_message?` implementation

bjpbakker:
By default `BaseCommandHandler#handles_message?` returns whether a
command is registered with the `on` DSL syntax of `SelfApplier`.

### DIFF
--- a/lib/sequent/core/base_command_handler.rb
+++ b/lib/sequent/core/base_command_handler.rb
@@ -26,6 +26,10 @@ module Sequent
         @repository = repository
       end
 
+      def handles_message?(command)
+        self.class.message_mapping.keys.find { |x| command.is_a? x }
+      end
+
       protected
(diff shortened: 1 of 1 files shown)


bforma:
:+1:

lib/sequent/core/base_command_handler.rb
```diff
@@ -26,6 +26,10 @@ def initialize(repository = Sequent.configuration.aggregate_repository)
         @repository = repository
       end
 
+      def handles_message?(command)
+        self.class.message_mapping.keys.find { |x| command.is_a? x }
```
erikrozendaal: This behavior seems to be inconsistent with https://github.com/zilverline/sequent/blob/improvement/default-handles-message/lib/sequent/core/helpers/self_applier.rb#L35

(This one allows subclasses while `SelfApplier` does not.

bjpbakker:
:+1: 

Fixed it on master.

PR #16, opened 2015-07-09
Support `update` on attribute support

bjpbakker:
Updating an object with attribute support will return a new instance of
such object with the given changes merged in.


bforma:
spec?

bforma:
:+1:

PR #18, opened 2015-07-17
Spec helper improvements

bjpbakker:
Spec helpers `given_events` and `then_events` now can use both varargs and a list of events.

Also any unhandled errors are raised after a spec.

### DIFF
--- a/lib/sequent/test/command_handler_helpers.rb
+++ b/lib/sequent/test/command_handler_helpers.rb
@@ -128,17 +128,17 @@ module Sequent
         begin
           @command_handler.handle_message(command)
         rescue => e
-          @actual_error = e
+          @unhandled_error = e
         end
         @repository.commit(command)
       end
 
       def then_fails_with clazz
-        expect(@actual_error).to be_kind_of(clazz)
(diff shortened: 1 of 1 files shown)


lib/sequent/test/command_handler_helpers.rb
```diff
@@ -116,33 +116,31 @@ def deserialize_events(events)
             Class.const_get(type).deserialize_from_json(Sequent::Core::Oj.strict_load(json))
           end
         end
-
       end
 
       def given_events *events
-        @event_store.given_events(events)
+        @event_store.given_events(events.flatten(1))
```
derekkraan: You can already do this: `given_events(event1, event2, event3)`

bjpbakker: It's because we want to do `given_events(events_list1 + events_list2)` with no more parentheses and splats :smiley_cat:

derekkraan: :+1:

s0meone: :+1:

bforma:
:+1:

s0meone:
:+1:

PR #19, opened 2015-07-17
Fix warnings about `raise_exception` in specs

s0meone:
Fixes these kind of warnings:

> WARNING: Using the `raise_error` matcher without providing a specific error or message risks false positives, since `raise_error` will match when Ruby raises a `NoMethodError`, `NameError` or `ArgumentError`, potentially allowing the expectation to pass without even executing the method you are intending to call. Instead consider providing a specific error class or message. This message can be supressed by setting: `RSpec::Expectations.configuration.warn_about_potential_false_positives = false`. Called from /Users/daniel/Development/Workspaces/code/sequent/spec/lib/sequent/core/aggregate_repository_spec.rb:91:in `block (2 levels) in <top (required)>'.

### DIFF
--- a/spec/lib/sequent/core/aggregate_repository_spec.rb
+++ b/spec/lib/sequent/core/aggregate_repository_spec.rb
@@ -88,7 +88,7 @@ describe Sequent::Core::AggregateRepository do
   end
 
   it "should raise exception if a aggregate does not exists" do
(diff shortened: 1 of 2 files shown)


bforma:
:+1:

bjpbakker:
:+1:

PR #23, opened 2015-07-23
Workflow spec helpers

bjpbakker:
Provide some spec helpers for testing workflows.

Also fix kind of an important bug in `Workflow`.


bforma:
:+1:

PR #24, opened 2015-08-03
Do not raise ActiveRecord::RecordNotFound

lvonk:
Before StreamRecord existed the EvenStore raised an AggregateNotFound when it was not found in the EventStore.


bjpbakker:
:+1:

PR #25, opened 2015-08-03
Remove dsl for exception checking

lvonk:
The then_fails_with limits the options of checking the raised exception (only on class). 
Also we want to avoid Death-By-DSL. For instance rspec alreay provides a nice way to check for exceptions.


lvonk:
Hi @acarpe,

We are going to remove the `then_fails_with` from sequent for reasons stated above.

PR #29, opened 2015-08-14
Add test helper (patch) for datetime comparisons

bforma:
This patch on `Time`, `DateTime` and `ActiveSupport::TimeWithZone` omits nanosecond difference when comparing times. 

Allows you to keep using built-in rspec matchers (such as `eq`) without having to worry about nanosecond differences which are likely to occur when running on a Linux machine or when time objects have been serialized (ie. Postgres) and lost subsecond precision.

Usage:

Just `require 'sequent/test/time_comparision'` in your spec file (or spec_helper.rb), and you're good to go.

### DIFF
--- a/lib/sequent/test/time_comparison.rb
+++ b/lib/sequent/test/time_comparison.rb
@@ -1,0 +1,40 @@
+module Sequent
+  module Test
+    module DateTimePatches
+      module Normalize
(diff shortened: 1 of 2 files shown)


lvonk:
:+1:

PR #30, opened 2015-08-18
Extend support for parsing arrays from params hash

lvonk:
This commit also enables parsing array(Integer) and other simple types.
Furthermore the as_params, used mostly in specs, now also creates
similar hashes for posting as your browser would do.

### DIFF
--- a/lib/sequent/core/helpers/param_support.rb
+++ b/lib/sequent/core/helpers/param_support.rb
@@ -30,7 +30,13 @@ module Sequent
             if type.respond_to? :from_params
               value = type.from_params(value)
             elsif type.is_a? Sequent::Core::Helpers::ArrayWithType
-              value = value.map { |v| type.item_type.from_params(v) }
+              value = value.map do |v|
+                if type.item_type.respond_to?(:from_params)
+                  type.item_type.from_params(v)
+                else
+                  v
+                end
+              end
             end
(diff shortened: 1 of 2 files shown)


lib/sequent/core/helpers/param_support.rb
```diff
@@ -45,36 +51,54 @@ def as_params
...
-              value = value.as_params
-            elsif value.kind_of?(Array)
-              value = value.map { |val| val.kind_of?(ValueObject) ? val.as_params : val }
-            elsif value.is_a? DateTime
-              value = value.iso8601
-            elsif value.is_a? Date
-              value = value.strftime("%d-%m-%Y") # TODO Remove to TypeConverter
-            end
-            hash[field[0]] = value
+            hash[field[0]] = if value.kind_of?(Array)
```
bjpbakker: Should we support any `Enumerable` here, not just `Array`?

bjpbakker:
Nice refactoring :+1:

lib/sequent/core/helpers/param_support.rb
lvonk: I think we eventually should, this would I guess be another PR since we have to look into ArrayWithType and array(Integer) etc.

bjpbakker: Agreed.

PR #33, opened 2015-10-08
Fix auto-index on aggregate_id

derekkraan:
The bug was that only one record was being saved per aggregate_id. This
makes no sense.


lvonk:
:+1:

PR #34, opened 2015-10-09
Show more info on an error during replay

bjpbakker:


lib/sequent/core/event_store.rb
```diff
@@ -108,7 +108,12 @@ def resolve_event_type(event_type)
...
         events.each do |event|
           event_handlers.each do |handler|
-            handler.handle_message event
+            begin
+              handler.handle_message event
+            rescue => e
+              msg = "Replay failed => aggregate: #{event.aggregate_id}, event: #{event.class}, sequence_number: #{event.sequence_number}"
+              raise e, msg, e.backtrace
+            end
           end
```
lvonk: I suggest a named Error like `ReplayFailedError` with the `EventHandler`s / `Projector`s class name, the Original exception and the full event. Than the caller can make up it's own message.

bjpbakker:
Great feedback, agreed. I went with `PublishEventError` as the error handling isn't just related to replaying.

The error now includes the `event_handler_class` and the `event` as well as the `cause` (which is automatically set by Ruby anyway).

lvonk:
Cool, I didn't know about `cause`...

lvonk:
:+1:

PR #37, opened 2015-10-21
Provide Rake tasks to ease setup

bjpbakker:
Make Sequent setup a bit less intrusive with some rake tasks to create and maintain the database and the required schemas (for the event store and the view projections).

### DIFF
--- a/lib/sequent/rake/tasks.rb
+++ b/lib/sequent/rake/tasks.rb
@@ -95,7 +95,7 @@ module Sequent
       end
 
       def view_projection
-        @view_projection ||= Sequent::Support::ViewProjection.new(options.fetch(:view_schema))
+        options.fetch(:view_projection)
       end
     end
   end
--- a/lib/sequent/rake/tasks.rb
+++ b/lib/sequent/rake/tasks.rb
@@ -1,0 +1,104 @@
+require 'active_record'
+require 'rake'
+require 'rake/tasklib'
+
+require 'sequent/support'
+
+module Sequent
+  module Rake
+    class Tasks < ::Rake::TaskLib
+      include ::Rake::DSL
+
+      DEFAULT_OPTIONS = {
+        migrations_path: 'db/migrate',
+        event_store_schema: 'public'
+      }
+
+      attr_reader :options
+
+      def initialize(options)
+        @options = DEFAULT_OPTIONS.merge(options)
+      end
+
+      def register!
+        register_db_tasks!
+        register_view_schema_tasks!
+      end
+
+      def register_db_tasks!
+        namespace :db do
+          desc 'Create the database'
+          task :create do
+            current_environments.each do |env|
+              env_db = db_config(env)
+              puts "Create database #{env_db['database']}"
+              Sequent::Support::Database.create!(env_db)
+            end
+          end
+
+          desc 'Drop the database'
+          task :drop do
+            current_environments.each do |env|
+              env_db = db_config(env)
+              puts "Drop database #{env_db['database']}"
+              Sequent::Support::Database.drop!(env_db)
+            end
+          end
+
+          task :establish_connection do
+            env_db = db_config(options.fetch(:environment))
+            Sequent::Support::Database.establish_connections(env_db, view_projection.schema_name)
+          end
+
+          desc 'Migrate the database'
+          task migrate: :establish_connection do
+            database.create_schema_if_not_exists!(options.fetch(:event_store_schema))
+            database.migrate(options.fetch(:migrations_path))
+          end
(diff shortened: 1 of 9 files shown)


lib/sequent/support/database.rb
```diff
@@ -0,0 +1,63 @@
...
+      def create_schema!(schema)
+        sql = "CREATE SCHEMA #{schema}"
+        if user = ActiveRecord::Base.connection_config[:username]
+          sql += " AUTHORIZATION #{user}"
+        end
+        ActiveRecord::Base.connection.execute(sql)
+      end
+
+      def create_schema_if_not_exists!(schema)
+        create_schema!(schema) unless schema_exists?(schema)
```
lvonk: `CREATE SCHEMA IF NOT EXISTS`?

lib/sequent/support/database.rb
```diff
@@ -0,0 +1,63 @@
...
+        ActiveRecord::Base.connection.create_database(db_config['database'])
+      end
+
+      def self.drop!(db_config)
+        ActiveRecord::Base.establish_connection(db_config.merge('database' => 'postgres'))
+        ActiveRecord::Base.connection.drop_database(db_config['database'])
+      end
+
+      def self.establish_connections(db_config, view_schema_name)
+        Sequent::Support::BaseViewModel.establish_connection(
```
lvonk: This won't work imho since in this setup the BaseViewModel and event store classes have different connection pools. 
Then we can no longer commit and update to the view model and event store in the same transaction, because they use different connections.
This is fine we we **only** want to support eventual consistency, but that is not the case.

lib/sequent/rake/tasks.rb
```diff
@@ -0,0 +1,104 @@
...
+            end
+          end
+
+          task :establish_connection do
+            env_db = db_config(options.fetch(:environment))
+            Sequent::Support::Database.establish_connections(env_db, view_projection.schema_name)
+          end
+
+          desc 'Migrate the database'
+          task migrate: :establish_connection do
```
lvonk: Why have this if we don't have migrations for the view_schema?

lib/sequent/rake/tasks.rb
```diff
@@ -0,0 +1,104 @@
...
+            database.drop_schema!(view_projection.schema_name)
+          end
+        end
+      end
+
+      private
+
+      def current_environments
+        envs = [options.fetch(:environment)]
+        envs << 'test' if @env == 'development'
```
lvonk: Why?

bjpbakker: I wondered about this too for a while. It's what Rails does. It is helpful to maintain the test database without having to set an environment (so that after `db:create` and `db:migrate` ActiveRecord can be used in a spec helper to maintain the test schema).

If it's only confusing, we should remove it :)

lib/sequent/rake/tasks.rb
bjpbakker: These migrations should migrate the event store schema, which should be changed through migrations.

Although we don't prevent anyone from migrating the view, migrations should only target the event store.

lib/sequent/support/database.rb
bjpbakker: As far as I can tell this would only affect consistency when view models influence consistency of the domain, which I think they shouldn't?

That said, it might be better to provide more control or configuration on this.

lib/sequent/support/database.rb
bjpbakker: :+1:

lib/sequent/support/database.rb
lvonk: What I mean is that currently in a single transaction the events are stored in the event store and the view is updated in the projections. When you use 2 different connectionpools (one for the projections, and one for the event store) this is no longer possible.

So for instance when the event store fails to update, for instance concurrent updates of aggregates, then the view model is not rolled back because it has its own transaction. We experienced this issue with kaiwa (ask @s0meone)

lib/sequent/rake/tasks.rb
lvonk: :+1:

lib/sequent/support/database.rb
bjpbakker: I see, good point (and thanks for explaining :))

Will fix this by getting back to a single connection pool, and use a special connection to build the view schema (as it must be explicitly created in the database view schema).

s0meone: We had this exact issue in Kaiwa. @lvonk hits the nail right on the head :wink:

lvonk:
:+1:

PR #38, opened 2015-10-29
Support exotic column types by always casting

derekkraan:
replay of production data: `278.2s` with master, and `282.8s` with this change.


lvonk:
Nice. :+1:

PR #39, opened 2015-11-03
Remove explicit support for multitenancy

derekkraan:
cc @lvonk and @s0meone 

If you are using these in your own app, copy + paste the code removed here and include in your own project.

le sigh ...
screen shot 2015-11-03 at 14 11 16


s0meone:
No, not using these classes. Thanks for the heads up. Glad these will be removed though :+1:

bjpbakker:
Nice 👍

Please update the `db/sequent_schema.rb` and remove the multi tenancy fields there too :)

lvonk:
:shipit:

PR #41, opened 2015-11-03
Fix type checking array params in `from_params`

bjpbakker:
Delegate the type checking for array types to the association validator,
while type checking the elements of the array in param support (similar
to any other param).

### DIFF
--- a/lib/sequent/core/helpers/array_with_type.rb
+++ b/lib/sequent/core/helpers/array_with_type.rb
@@ -16,6 +16,10 @@ module Sequent
         def to_s
           "Sequent::Core::Helpers::ArrayWithType.new(#{item_type})"
         end
+
+        def candidate?(value)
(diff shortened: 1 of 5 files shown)


lvonk:
I thought we would raise instead of creating invalid objects?

bjpbakker:
In the 'raw' type conversion from strings we raise indeed. Then it turned out that param support is doing some type checking itself, including deed array inspection. For array types with a non-array value this crashed too.

So This PR fixes that to be consistent with other param type checks (like integer).

lvonk:
:+1:

derekkraan:
This breaks some functionality in Freemle. Not 100% sure how to fix, but it turns out that in Freemle we sometimes have a type `String` but a value `["foo", "bar", "baz"]`. (in `LedgerAccount`)

PR #43, opened 2015-11-25
Freemle improvements

derekkraan:
This is going to cause quite a few deprecation errors, @lvonk @s0meone @bjpbakker. (Not for Freemle though; I already switched over to `Sequent.new_uuid` there)


lvonk:
:+1:

PR #44, opened 2015-11-30
Encode BigDecimal as string and decode as BigDecimal

derekkraan:


bjpbakker:
Specs?

derekkraan:
Specs!

lvonk:
:shipit:

PR #45, opened 2015-11-30
Fix reading 'blank' values from params

bjpbakker:
Values like `false`, `[]`, `''` are `blank` according to ActiveRecord.
Though these represent actual values that you need to deal with. In
params there's a difference between not-present and blank values. Some
of them might be dealt with in the same way (e.g. the params are
invalid), others might be actual values (like false for a Boolean).


derekkraan:
:shipit:

PR #47, opened 2015-12-01
Change serialising of events to use as_json directly

lvonk:
Previously `event.attributes` was used to serialise events. 
Due to this it was not really possible to change the way an object is
serialised to and from JSON.

By introducing `as_json` explicitly for serialising this option becomes
available.

### DIFF
--- a/lib/sequent/core/event_store.rb
+++ b/lib/sequent/core/event_store.rb
@@ -39,7 +39,7 @@ module Sequent
       #
       def commit_events(command, streams_with_events)
         store_events(command, streams_with_events)
-        publish_events(streams_with_events.flat_map {|_, events| events}, event_handlers)
+        publish_events(streams_with_events.flat_map { |_, events| events }, event_handlers)
       end
 
       ##
@@ -72,7 +72,7 @@ SELECT event_type, event_json
       #
       # @param block that returns the events.
       def replay_events
-        events = yield.map {|event_hash| deserialize_event(event_hash)}
+        events = yield.map { |event_hash| deserialize_event(event_hash) }
         publish_events(events, event_handlers)
       end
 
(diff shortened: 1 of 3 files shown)


lvonk:
@bjpbakker @s0meone Please check. Events are no longer serialised via `event.attributes` but via `event.as_json` which adheres more to the `Oj` way of serialising and deserialising. `Oj`s order is:

- `to_hash`
- `as_json`
- `to_json`

By providing `as_json` on `AttributeSupport` which uses `attributes` internally the result should be the same.

lvonk:
Ping @s0meone @bjpbakker

lvonk:
This PR will auto-merge in 1 day... @s0meone @bjpbakker

lib/sequent/core/event_store.rb
```diff
@@ -140,7 +140,16 @@ def store_events(command, streams_with_events = [])
...
           uncommitted_events.each do |event|
-            event_record_class.create!(command_record: command_record, stream_record_id: event_stream.stream_record_id, event: event)
+            values = {command_record: command_record,
+                      stream_record_id: event_stream.stream_record_id,
+                      aggregate_id: event.aggregate_id,
+                      sequence_number: event.sequence_number,
+                      event_type: event.class.name,
+                      event_json: event_record_class.serialize_to_json(event),
+                      created_at: event.created_at}
+            values = values.merge(organization_id: event.organization_id) if event.respond_to?(:organization_id)
```
bjpbakker: Wasn't `organization_id` removed by @derekkraan in an earlier PR?

bjpbakker:
We should probably document that `#to_hash` is used for JSON serialization if both `#to_hash` and `#as_json` are implemented. For me that would be totally unexpected.

bjpbakker:
Other than that, very nice PR! :+1:

lib/sequent/core/event_store.rb
lvonk: Yes it was... Let me see how to fix.

lvonk:
Will change organization_id later

PR #49, opened 2015-12-09
Improve `then_events`.

derekkraan:
Now you can test your events like this:

```
event1 = Event.new
event3 = Event.new
then_events event1, OtherEvent, event3
```

Then Sequent will check that the resulting events are of type `[Event,
OtherEvent, Event]`, and will also check the attributes of `event1` and
`event3` against the first and third events.

### DIFF
--- a/lib/sequent/test/command_handler_helpers.rb
+++ b/lib/sequent/test/command_handler_helpers.rb
@@ -125,9 +125,12 @@ module Sequent
         @repository.clear
       end
 
-      def then_events *events
-        expect(@event_store.stored_events.map(&:class)).to eq(events.flatten(1).map(&:class))
-        @event_store.stored_events.zip(events.flatten(1)).each do |actual, expected|
+      def then_events(*expected_events)
+        expected_classes = expected_events.flatten(1).map { |event| event.class == Class ? event : event.class }
+        expect(@event_store.stored_events.map(&:class)).to eq(expected_classes)
+
+        @event_store.stored_events.zip(expected_events.flatten(1)).each do |actual, expected|
+          next if expected.class == Class
           expect(Sequent::Core::Oj.strict_load(Sequent::Core::Oj.dump(actual.payload))).to eq(Sequent::Core::Oj.strict_load(Sequent::Core::Oj.dump(expected.payload))) if expected
         end
       end
--- a/lib/sequent/test/event_handler_helpers.rb
+++ b/lib/sequent/test/event_handler_helpers.rb
@@ -39,9 +39,12 @@ module Sequent
         end
       end
 
-      def then_events *events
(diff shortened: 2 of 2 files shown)


lvonk:
Strange that we have the same code. Bu other than that: :shipit:

derekkraan:
anyone else want to chime in? @bjpbakker @s0meone

lib/sequent/test/command_handler_helpers.rb
```diff
@@ -125,9 +125,12 @@ def when_command command
         @repository.clear
       end
 
-      def then_events *events
-        expect(@event_store.stored_events.map(&:class)).to eq(events.flatten(1).map(&:class))
-        @event_store.stored_events.zip(events.flatten(1)).each do |actual, expected|
+      def then_events(*expected_events)
+        expected_classes = expected_events.flatten(1).map { |event| event.class == Class ? event : event.class }
```
bjpbakker: `event === Class` ?

bjpbakker:
Nice, although we tend use a custom `then_events` in projects that separates event classes from payloads (so we can partial match payloads).

Maybe we need to submit a PR for that too.. :smirk:

:shipit:

lib/sequent/test/command_handler_helpers.rb
derekkraan: What would that do? I know what `Class === event` does, not sure what `event === Class` does.

lvonk:
Partial match payloads seems a nice feature! Bring the PR's on.

lib/sequent/test/command_handler_helpers.rb
```diff
@@ -125,9 +125,12 @@ def when_command command
...
+        expect(@event_store.stored_events.map(&:class)).to eq(expected_classes)
+
+        @event_store.stored_events.zip(expected_events.flatten(1)).each do |actual, expected|
```
lvonk: Can't we just use `flat_map` btw? Why explicit `flatten(1)`?

lib/sequent/test/command_handler_helpers.rb
lvonk: Can't we just use `flat_map` btw? Why explicit `flatten(1)`? I know it was already there, but I don't remember why it was put in like this

bjpbakker: Because there's only `flatten(1)` and no `map`

bjpbakker: @derekkraan normally I would suspect `A === a` and `a === A` to do the same thing, but then again it might be that `A` provides its own `===` implementation.

Maybe go with `Class === event` instead of `event.class == Class`?

derekkraan:
I'm merging this one cause we need it for Freemle. Improvements welcome as PRs :)

PR #51, opened 2016-01-06
Add specific error in case event_json is messed up

lvonk:
Instead of only receiving an implicit conversion error,
this also adds the actual event that caused that error.


derekkraan:
:shipit:

PR #52, opened 2016-01-21
Order events for replay by stream and sequence

bjpbakker:
Event migrations can cause events to be inserted in existing event
streams. When replaying events by their event ID causes these events to
be replayed at the end of the stream.

This patch changes the order of events for replay to the ID of the
stream and within a stream the event's sequence number. In most
situations the result should be identical. In cases where not, the
sequence number ordering is what you'd expect (since this is also how
Sequent loads an aggregate from the event store).

### DIFF
--- a/lib/sequent/support/view_projection.rb
+++ b/lib/sequent/support/view_projection.rb
@@ -15,7 +15,8 @@ module Sequent
 
           load schema_definition
           event_store = Sequent.configuration.event_store
-          event_store.replay_events { Events::ORDERED_BY_ID[event_store] }
+          ordering = Events::ORDERED_BY_STREAM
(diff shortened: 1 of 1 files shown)


lvonk:
I don't understand how this works? An aggregate seems to have only one Stream?

bjpbakker:
An aggregate has a single event stream. Within this stream, events should be ordered based on `sequence_number`.

By sorting all events by `stream.id` first and then by `sequence_number` we get all streams in order of creation, with the correct ordering of events within the stream. Without this, events that are inserted into the stream by a migration, are replayed last causing some projections to fail (e.g. a delete is replayed before an update).

lvonk:
:+1:

PR #54, opened 2016-02-15
Fix validation of inherited array properties

bjpbakker:
Include the superclass' types when finding the `type_for` an attribute.

### DIFF
--- a/lib/sequent/core/helpers/attribute_support.rb
+++ b/lib/sequent/core/helpers/attribute_support.rb
@@ -33,11 +33,6 @@ module Sequent
             end
           end
 
-          def type_for(name)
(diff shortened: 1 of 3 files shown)


lib/sequent/core/helpers/attribute_support.rb
```diff
@@ -34,7 +34,8 @@ def types
           end
 
           def type_for(name)
-            @types.find { |k, _| k == name }.last
+            _, type = types.find { |k, _| k == name }
```
lvonk: types[name]?

bjpbakker: Good point. Inline `type_for` as `types.fetch` now.

PR #56, opened 2016-02-25
Extend StandardError instead of Exception

remvee:
Extend 'Exception' for exceptions, unexpected external factors; network
down, out of memory etc.  Extend 'StandardError' to signal application
level errors.

Ruby's 'rescue', without supplying an exception type, will catch
'StandardError' and children.  Exceptions extending 'Exception' will
need to be stated in the rescue statement making it very hard for
frameworks supply 'recue_from' like features.


bforma:
Very good! :+1:

derekkraan:
agree

PR #57, opened 2016-03-08
Wrap ActiveRecord::RecordNotUnique with Sequent::Core::EventStore::OptimisticLockingError

derekkraan:
This will impact your app if you are handling `ActiveRecord::RecordNotUnique` originating from the event store.

@bjpbakker @s0meone @erikrozendaal comments?

### DIFF
--- a/lib/sequent/core/event_store.rb
+++ b/lib/sequent/core/event_store.rb
@@ -22,6 +22,14 @@ module Sequent
         end
       end
 
+      class OptimisticLockingError < RuntimeError
+        attr_reader :event
+
+        def initialize(event)
+          @event = event
+        end
+      end
+
       class DeserializeEventError < RuntimeError
         attr_reader :event_hash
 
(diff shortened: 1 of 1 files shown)


lvonk:
Ping @bjpbakker @s0meone. We want to wrap the RecordNotUnique so we can react to this. If you are using this somehow you will need to change this :-)

bjpbakker:
:+1:

lib/sequent/core/event_store.rb
```diff
@@ -164,11 +172,14 @@ def store_events(command, streams_with_events = [])
                       created_at: event.created_at}
             values = values.merge(organization_id: event.organization_id) if event.respond_to?(:organization_id)
 
-            event_record_class.create!(values)
+            begin
+              event_record_class.create!(values)
+            rescue ActiveRecord::RecordNotUnique => e
```
bjpbakker: `e` is unused

derekkraan: thanks, fixed!

PR #58, opened 2016-03-09
Domain errors

derekkraan:
This can be used in your domain models to raise *domain* exceptions.

Here's an example of how we use this in Freemle:

```
class Invoice < Sequent::Core::AggregateRoot
  class InvoiceAlreadySentError < DomainError; end

  def send
    fail InvoiceAlreadySentError.new(date: @sent_on_date) if @sent_on_date
  end
end
```

Then you can rescue DomainError in your base controller:

```
app.error Sequent::Core::AggregateRoot::DomainError do |e|
  flash[:error] = e.message
  redirect back
end
```

In your translations, you can use `date` as normal:

```
en:
  errors:
    Invoice::InvoiceAlreadySentError: 'This invoice was already sent on %{date}'
```

### DIFF
--- a/lib/sequent/core/aggregate_root.rb
+++ b/lib/sequent/core/aggregate_root.rb
@@ -37,6 +37,22 @@ module Sequent
 
       attr_reader :id, :uncommitted_events, :sequence_number, :event_stream
 
+      class DomainError < RuntimeError
+        attr_reader :i18n_parameters
+
+        def initialize(i18n_parameters = {})
+          @i18n_parameters = i18n_parameters
+        end
+
+        def message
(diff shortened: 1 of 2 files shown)


bjpbakker:
Not sure it's a good idea to have Sequent depend on i18n. It's easy enough to implement a simple `DomainError` class in your project, using your own conventions and your own dependencies.

lvonk:
Although we probably already depend on `i18n` via `activemodel` I agree with @bjpbakker. This can be dealt with in your/our own project.

derekkraan:
Ok, dan trek ik deze PR in.

PR #59, opened 2016-04-01
Insert all events per commit in a single statement

lvonk:
When a command can result in multiple events inserting
each event_record separately can become slow.
By using 

    insert into (..) values (..),(..),(..)

we save roundtrips to the database.

Possible downside is that when an ActiveRecord::RecordNotUnique
is thrown we no longer know which event caused the error.

### DIFF
--- a/lib/sequent/core/event_store.rb
+++ b/lib/sequent/core/event_store.rb
@@ -23,11 +23,6 @@ module Sequent
       end
 
       class OptimisticLockingError < RuntimeError
-        attr_reader :event
-
-        def initialize(event)
-          @event = event
-        end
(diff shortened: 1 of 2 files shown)


lvonk:
@derekkraan @s0meone @bjpbakker @erikrozendaal 

Thoughts? This saves a lot of inserts when doing bulk insert when for instance uploading a file generating lots (> 100) events.

derekkraan:
Sounds like a good idea to me. Let's deal with the "which event caused it to fail?" problem when it becomes one.

bjpbakker:
:shipit:

PR #60, opened 2016-05-30
Remove leading zeros from Integer values

lvonk:
We register validates_numericality_of when using Integer

The value ’09’ is considered numerical but is not a valid Integer 
according to Integer(’09’). So we strip leading zeros since they have
no meaning in Integers.

### DIFF
--- a/spec/lib/sequent/core/command_service_spec.rb
+++ b/spec/lib/sequent/core/command_service_spec.rb
@@ -83,5 +83,12 @@ describe Sequent::Core::CommandService do
                                                                 expect(e.errors[:value]).to eq ['is not a number']
                                                               end
     end
+
+    it "does not removes leading zeros when using hexadecimal values" do
+      command = WithIntegerCommand.new(aggregate_id: "1", value: "0x10")
+      expect { command_service.execute_commands(command) }.to raise_error do |e|
+                                                                expect(e.errors[:value]).to eq ['is not a number']
+                                                              end
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


erikrozendaal:
Leading zeroes do have meaning when using the `Integer` conversion method:

```ruby
irb(main):001:0> Integer('050')
=> 40
irb(main):002:0> Integer('0x30')
=> 48
```

`to_i` doesn't try to parse octal or hexadecimal numbers, but fails silently when passed a non-numeric string. Maybe a better conversion is to check that the input string consists of only decimal digits and then use `to_i`?

lvonk:
Thanks @erikrozendaal. We will go for base 10 than.

spec/lib/sequent/core/command_service_spec.rb
```diff
@@ -60,11 +60,28 @@ class WithIntegerCommand < Sequent::Core::BaseCommand
...
+
     it "does not parse values if the command is invalid" do
       command = WithIntegerCommand.new(value: "A")
       expect { command_service.execute_commands(command) }.to raise_error do |e|
                                                                 expect(e.errors[:value]).to eq ['is not a number']
                                                               end
     end
+
+    it "does not removes leading zeros if command is invalid" do
+      command = WithIntegerCommand.new(aggregate_id: "1", value: "0x")
```
erikrozendaal: Test with a valid hexadecimal number (eg `0x10`).

erikrozendaal:
:shipit:

PR #63, opened 2016-06-15
`then_commands` asserts that commands are valid

bjpbakker:
Executes commands should be valid. Since `then_commands` is a test helper that asserts the attributes of recorded commands (using an internal recorder setup), it should validate that executed commands are valid, too.


derekkraan:
:shipit: wmb

PR #65, opened 2016-06-30
Register subclasses of Sequent::Core::Command with

derekkraan:
Sequent::Core::Commands

This avoids the use of `String#constantize` when you want to find a
command from a string. The syntax is
`Sequent::Core::Commands.find(string)`.


lvonk:
:shipit:

PR #67, opened 2016-07-18
Update replay session indexes when updating a record

derekkraan:
Problem: when updating a record, using the replay session, the index wasn't being updated as well.

This caused problems when later trying to find a record based on the updated attribute (if that attribute was indexed). This is especially likely to have happened if using `update_all_records` or `delete_all_records`.


lvonk:
👍

PR #74, opened 2016-09-28
Nodowntime replay

lvonk:
Enable projects to replay events without downtime. The strategy is as follows:

1) Create new view schema
2) Replay all events
3) When done put the app in maintenance mode (this is only for a short time)
4) Replay events not yet replayed
5) Switch view schema
6) Restart your app

To keep track of events not yet replayed the `on_progress` block is passed in the ids of the
events replayed. The app that is deployed needs to keep track of these ids so in step 4 you can select all events `where id not in (replayed_ids)`.

### DIFF
--- a/lib/sequent/core/event_store.rb
+++ b/lib/sequent/core/event_store.rb
@@ -106,16 +106,21 @@ SELECT event_type, event_json
                                     on_progress: PRINT_PROGRESS)
         progress = 0
         cursor = get_events.call
+        ids_replayed = []
         cursor.each_row(block_size: block_size).each do |record|
           event = deserialize_event(record)
           publish_events([event], event_handlers)
           progress += 1
-          on_progress[progress, false] if progress % block_size == 0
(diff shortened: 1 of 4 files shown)


lib/sequent/core/event_store.rb
```diff
@@ -106,16 +106,21 @@ def replay_events_from_cursor(block_size: 2000,
...
           progress += 1
-          on_progress[progress, false] if progress % block_size == 0
+          ids_replayed << record['id']
+          if progress % block_size == 0
+            on_progress[progress, false, ids_replayed.compact]
+            ids_replayed.clear
+          end
         end
-        on_progress[progress, true]
+        on_progress[progress, true, ids_replayed.compact]
```
bjpbakker: Why compact? Every record should have an 'id' for this to work

lvonk:
@bjpbakker I will add that in the other PR

PR #77, opened 2016-10-26
Play all events per-event handler.

derekkraan:
Instead of replaying one event to all event handlers, and then moving on
to the next event. This caused a bug in workflows:

Say you append a number of events to an aggregate. The first of these
events triggers a workflow, which appends some extra events. These extra
events will be sent *first* to the event handlers, out of order.

To get the correct order of events being applied, it's still important to
always have workflows lower in the event_handlers list than projectors.


lvonk:
👍

PR #81, opened 2017-02-22
fixed: AggregateNotFound never got raised

derekkraan:


erikrozendaal:
The documentation for `load_aggregates` should match the implementation.

Maybe separate the error throwing version into a `!` method version? Or do you always expect an exception (contradiction really) when not finding the aggregate?

derekkraan:
The semantics from before `load_aggregate` was to always raise this exception when an `aggregate_id` was not found. This commit restores that behaviour.

I'll adjust the documentation.

erikrozendaal:
👍

PR #82, opened 2017-03-22
Add column names to index key to avoid clashes

lvonk:
It was possible to have clashes in indexed records when:

- having multiple indices on the same record;
- with the same number of indexed columns
- that could have the same value

For example:

An indexed record FooRecord with the following indices:

    FooRecord => [[:id, :val_1], [:id, :val_2]]

This was internally indexed with the key:

    [FooRecord, record[:id], record[:val_1]]

and

	[FooRecord, record[:id], record[:val_2]]

If val_1 and val_2 had the same value it used the same index key
resulting in unwanted behaviour (or view projection mayhem).

By adding the column names to the indexed key this now is truly unique:

    [FooRecord, :id, record[:id], :val_1, record[:val_1]]

and

    [FooRecord, :id, record[:id], :val_2, record[:val_2]]

### DIFF
--- a/lib/sequent/core/record_sessions/replay_events_session.rb
+++ b/lib/sequent/core/record_sessions/replay_events_session.rb
@@ -100,7 +100,10 @@ module Sequent
 
           def find(record_class, where_clause)
             key = [record_class.name]
-            get_index(record_class, where_clause).each { |field| key << where_clause[field] }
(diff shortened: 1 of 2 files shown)


lvonk:
Ping @derekkraan

derekkraan:
nice fix :shipit: :+1:

PR #83, opened 2017-05-11
Add `.handles_message?` method

derekkraan:
I have added this because it's a convenient interface that I can
overwrite when my own implementation of `.handle_message` differs.

It's similar to `.method_missing` and `.respond_to?`.

### DIFF
--- a/lib/sequent/core/aggregate_snapshotter.rb
+++ b/lib/sequent/core/aggregate_snapshotter.rb
@@ -6,7 +6,7 @@ module Sequent
 
     class AggregateSnapshotter < BaseCommandHandler
 
-      def handles_message?(message)
+      def self.handles_message?(message)
         message.is_a? SnapshotCommand
       end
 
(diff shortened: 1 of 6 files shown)


bjpbakker:
This method is already implemented in `BaseCommandHandler`, which includes `SelfApplier`.

Is the intension of this PR to /move/ `handles_message?` to `SelfApplier`? In which case we want to remove it from `BaseCommandHandler`.

I would agree that it makes more sense to have the implementation in `SelfApplier`.

lvonk:
I agree with @bjpbakker we can better move the `handles_message?` to the SelfApplier

derekkraan:
@s0meone @bjpbakker I've moved `#handles_message?` to `SelfApplier`, but only as a class method. If you've overwritten `#handles_message?` in a command handler in your own code (this should be rare), this will now be broken.

s0meone:
:+1:

lvonk:
👍

PR #93, opened 2017-07-25
Configure test environment to better reflect reality

derekkraan:


derekkraan:
bezwaar @lvonk @s0meone @erikrozendaal ?

PR #97, opened 2017-10-30
Add queue-based event pub to make sure events are published in order

derekkraan:
Run the included spec on `master` to see what happens without the new `EventPublisher`


lib/sequent/core/event_publisher.rb
```diff
@@ -0,0 +1,44 @@
...
+        process_events
+      end
+
+      def process_events
+        while(!@events_queue.empty?) do
+          event = @events_queue.pop
+          configuration.event_handlers.each do |handler|
+            begin
+              handler.handle_message event
+            rescue
```
lvonk: This rescue can be moved down, this makes the code a bit more readable

```
def process_events
rescue
end
```

derekkraan: Dat kan niet: dan hebben we geen referentie naar `event` meer.

lvonk: Doh, you are right

lib/sequent/core/event_publisher.rb
```diff
@@ -0,0 +1,68 @@
...
+        process_events
+      end
+
+      private
+
+      def events_queue
+        Thread.current[:events_queue] ||= Queue.new
+      end
+
+      def skip_if_locked(&block)
```
lvonk: Maybe a better name is `skip_if_publishing` or `skip_if_not_main_thread`

lib/sequent/core/event_publisher.rb
```diff
@@ -0,0 +1,57 @@
+module Sequent
+  module Core
+    #
+    # EventPublisher ensures that, for every thread, events will be published in the order in which they are meant to be published.
```
lvonk: I think "... will be published breadth (or width) first" is better.

To clarify, with breadth first I mean: first all first level events than the consequences of those events.

lib/sequent/core/event_publisher.rb
```diff
@@ -0,0 +1,57 @@
...
+    #
+    # This potentially introduces a wrinkle into your plans: You therefore should not split a "unit of work" across multiple threads.
+    #
+    # If you do not want this, you are free to implement your own version of EventPublisher and configure sequent to use it.
+    #
+    class EventPublisher
```
lvonk: I am thinking to split this in 2 concerns. This might help in understanding the concepts better and might be easier to extend with other strategies. This also allows us to support the current behavior and be backward compatible.

1) The Events queue with is basically `process_events`
2) The FIFO strategy" which is basically `events.each { |event| events_queue.push(event) }`. 

When someone want LIFO (or depth first) the event should be prepended to the Queue.

lib/sequent/core/event_publisher.rb
```diff
@@ -0,0 +1,57 @@
...
+      def publish_events(events)
+        return if configuration.disable_event_handlers
+        events.each { |event| events_queue.push(event) }
+        process_events
+      end
+
+      private
+
+      def events_queue
+        Thread.current[:events_queue] ||= Queue.new
```
lvonk: Queue only supports FIFO I think. We can safely use Array here to support both FIFO and LIFO.

lib/sequent/core/command_service.rb
```diff
@@ -34,21 +34,10 @@ def initialize(configuration)
...
-
-              raise CommandNotValid.new(command) unless command.valid?
-              parsed_command = command.parse_attrs_to_correct_types
-              command_handlers.select { |h| h.class.handles_message?(parsed_command) }.each { |h| h.handle_message parsed_command }
-              repository.commit(parsed_command)
-            end
-          end
-        ensure
-          repository.clear
+        commands.each do |command|
```
lvonk: Can be a one-liner: `each. {}`

lib/sequent/core/command_service.rb
```diff
@@ -57,6 +46,31 @@ def remove_event_handler(clazz)
 
       private
 
+      def process_commands
```
lvonk: Can we extract this to a separate class like the EventPublisher. Both are quite similar in intent and behavior. Same suggestion made for the EventPublisher also apply here then :-)

lib/sequent/core/event_publisher.rb
derekkraan: I can't imagine why we would want to support LIFO.

lib/sequent/core/command_service.rb
derekkraan: `CommandService` is a separate class like `EventPublisher` already.

lib/sequent/core/event_publisher.rb
derekkraan: I think we should hold off on implementing something like this until someone asks for it / needs it.

lib/sequent/core/event_publisher.rb
lvonk: We are changing the current behavior so I think it is good to support both cases and the current one as default, with maybe a warning message.

derekkraan: Neither of those scenarios is the current one.

lvonk: Yes you are right... Okay, let's communicate this in the changelog when we release.

derekkraan:
@erikrozendaal @s0meone @bjpbakker feedback? We will be merging this shortly.

bjpbakker:
I do like this change, it solves a bunch of problems with published event ordering.

That said, current behavior is changed quite a lot, and may actually break current implementations. Most likely the tests won't cover this (because often they simply assert published events). That makes it tough to keep production code bases that are not in active development up to date with sequent.

Can we provide the current behavior via a configuration option?

derekkraan:
This behaviour is configurable to a certain degree: One is free to simply implement her own version of `EventPublisher` and configure Sequent to use that with `Sequent.configuration.event_publisher = YourEventPublisher`. This is a couple of steps beyond trivial, but I feel that it's not an inordinate amount of work.

I'd rather we not officially support the legacy behaviour and the code that goes along with it.

If we get questions about this I'd be more than happy to help people gut `EventPublisher` and `CommandService` to suit their needs.

derekkraan:
I think also that we should release this with a minor level version number increase to indicate that there is new functionality being introduces. This would be version 1.1.0.

bjpbakker:
If we choose to no longer support the current order of event publishing (and stick with [semantic versioning](http://semver.org/)) we should bump to 2.0. This change currently is not backwards compatible.

derekkraan:
That depends on your definition of "backwards compatible". Every change is potentially backwards incompatible, yet we don't increment major version number for every single bugfix.

The question is whether the new ordering is compatible with the old ordering. The answer is: that depends on how our users have implemented their systems and to what degree they are dependent on the old ordering. I suspect that most of them are not, making it for most (or all) users a backwards compatible change.

Or do you suspect that this will break functionality for a non-trivial percentage of Sequent users?

erikrozendaal:
Running additional (event handler) generated commands in the same transaction may lead to huge, complicated transactions. This may not be what we want. But I guess handling new commands in separate transactions requires a lot of additional infrastructure (persistent queues, background jobs, etc). 

So I guess this is fine, but I think we should discourage people from generating new commands in event handlers for immediate execution.

derekkraan:
@bjpbakker let's talk about version numbers on Friday. Merging this since that isn't a blocking issue.

PR #106, opened 2018-04-20
Project generator

stephanvd:


lib/sequent/generator.rb
```diff
@@ -0,0 +1,47 @@
...
+      make_directory
+      copy_files
+      rename_app_file
+      replace_app_name
+    end
+
+    private
+
+    def make_directory
+      FileUtils.mkdir_p(name)
```
lvonk: should we scope this on current_dir. So e.g. `/` etc is allowed?

lib/sequent/generator/template_project/db/migrate/1_add_event_store.rb
```diff
@@ -0,0 +1,54 @@
+class AddEventStore < ActiveRecord::Migration[4.2]
```
lvonk: Why is this `4.2`? Also this PR don't test against `4.2` anymore, see travis.yml.

lib/sequent/generator/template_project/lib/account/commands.rb
```diff
@@ -0,0 +1,6 @@
+class Account
```
lvonk: Shouldn't this be a `module`?

sequent.gemspec
```diff
@@ -9,6 +9,7 @@ Gem::Specification.new do |s|
   s.authors     = ["Lars Vonk", "Bob Forma", "Erik Rozendaal"]
```
lvonk: `svandiepen` :-)

spec/lib/sequent/generator_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+require 'fileutils'
+
+describe Sequent::Generator do
+  let(:tmp_path) { 'tmp/sequent-generator-spec' }
+
+  around do |example|
+    FileUtils.rmtree(tmp_path)
+    FileUtils.mkdir_p(tmp_path)
+    Dir.chdir(tmp_path) { example.run }
+    # FileUtils.rmtree(tmp_path)
```
lvonk: Comment

lib/sequent/generator/template_project/lib/account/commands.rb
stephanvd: We reopen the class from lib/account.rb here.

PR #108, opened 2018-05-11
Add new view migrations strategy

lvonk:
The current way of migrating is to rebuild the entire view schema
when doing a version upgrade. This is fine if you don’t have many events
and you are okay with downtime.
When your event store is getting bigger replaying all events into
a new view schema takes too long and consumes too much space.

The Sequent::ViewSchema::Migrator migrates the view schema with
minimal downtime. Main method are #migrate_online and #migrate_offline.

It basically works as follows:

1) There is only one view schema
2) It checks which Projectors to replay and creates new tables for
these Projectors (named table_name_VERSION_SUFFIX)
3) In a separate process it replays all events and stores them in
the new table
4) It keeps track of all Events that are already replayed
5) When done shutdown your app
6) Now replay all events that haven’t been replayed yet
7) Start your application

The last part (renaming the tables from _VERSION_SUFFIX 
to without _VERSION_SUFFIX) happens in a transaction. So if anything
fails the old status quo is preserved and no upgrade has taken place.

- [ ] Add rake task for `migration_online`
- [ ] Add rake task for `migration_offline`
- [ ] Add documentation


lvonk:
Renamed branch to #110 sequent-30

PR #110, opened 2018-05-25
Sequent 30

lvonk:
New major version of Sequent. 

See `CHANGELOG.md` for an overview of all the changes.

Most a notable changes:

- New strategy for rebuilding projections
- New documentation site available on www.sequent.io
- Removed support for AR < 5
- Removed support for older rubies


bforma:
Fun fact: You can use dots in your `git branch` names!

PR #121, opened 2018-06-15
Guide: "Modelling the domain"

stephanvd:
- [x] Split getting started guides to different chapters.
- [x] Seperate tutorial of generated aggregate from adding event.
- [x] Link to concept guides when first mentioning concept.
- [ ] Start with writing an integration spec to "begin with the end in mind"

### DIFF
--- a/docs/docs/modelling-the-domain.md
+++ b/docs/docs/modelling-the-domain.md
@@ -55,6 +55,7 @@ end
 We only need the `publication_date` attribute. Commands always target an aggregate, so we already know what to change by its `aggregate_id`. We could set a `publish` flag, but the event already communicates this intent.
 
 _Learn all about commands in the [Command](/docs/concepts/command.html) concept guide._
+{: .notice}
 
 ### Handling our new command
 
@@ -83,6 +84,7 @@ end
 Sequent retrieves the post for us and we call the (to be defined) `publish` method on the returned `Post` instance.
 
 _Learn all about command handlers in the [CommandHandler](/docs/concepts/command-handler.html) concept guide._
+{: .notice}
 
 ### The Aggregate Root
 
@@ -112,3 +114,4 @@ end
 ```
 
 _Learn all about aggregate roots in the [AggregateRoot](/docs/concepts/aggregate-root.html) concept guide._
+{: .notice}
--- a/docs/docs/modelling-the-domain.md
+++ b/docs/docs/modelling-the-domain.md
@@ -4,7 +4,7 @@ title: Modelling your Domain in Sequent
 
 ## A tour of the `Post` Aggregate
 
-The app we generated in [Getting Started](/docs/getting-started.html) comes with an example `Post` aggregate. In this guide we will take a quick look at Sequent's directory structure, go over some of the concepts used in Sequent by expanding on `Post` and create our very own `Comment` aggregate.
+The app we generated in [Getting Started](/docs/getting-started.html) comes with an example `Post` aggregate. In this guide we will take a quick look at Sequent's directory structure, go over some of the concepts used in Sequent by expanding on `Post` and create our very own `Author` aggregate.
 
 ### Directory structure
 
@@ -71,6 +71,8 @@ end
 
 Because we are adding a new aggregate `add_aggregate` is called with `Post.new(command)` as its argument. The actual business logic of what `Post` looks like is contained in the aggregate root.
 
+You are free to define your own signature of the constructor. In the example we chose to pass the command as argument, but nothing prevents you to define it using the separate attributes.
+
 We can add our own `on` block below the `AddPost` one:
 
 ```ruby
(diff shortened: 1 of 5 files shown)


docs/docs/modelling-the-domain.md
```diff
@@ -4,12 +4,114 @@ title: Modelling your Domain in Sequent
...
+{: .notice}
 
 ### The Aggregate Root
 
-In `lib/post/post.rb` we find the aggregate root.
+In `lib/post/post.rb` we find the aggregate root. This class encapsulates your business logic. Events are applied to instances of `Post` to give it its current state. We can see here that create a new `Post` will apply multiple events. Besides `PostAdded` we're also applying events to change the author, title and content. You might be tempted to group all those fields in one event, which can be a good idea if those fields always change together. We're using multiple events to emphasise that a single command does not alway correlate to a single event.
+
+```ruby
+class Post < Sequent::AggregateRoot
+  def initialize(command)
```
mattgibson: What happens here when an aggregate is reconstructed from the event stream? Does the command somehow get passed in again automatically? Bit unclear about how this method is used later in the lifecycle when `PostAdded` has already happened and we are now handling `PostUpdated` or similar.

docs/docs/modelling-the-domain.md
```diff
@@ -4,12 +4,114 @@ title: Modelling your Domain in Sequent
...
+  # ...
+end
+```
+
+Let's define how the domain should behave when receiving our new `PublishPost` command. Below the initialize method add:
+
+```ruby
+def publish(publication_date)
+  fail PostAlreadyPubishedError if @publication_date.any?
+  apply PostAdded
```
mattgibson: I'm confused as to why `PostAdded` gets applied here. Is this correct? I was expecting a `PostPublished` event.

docs/docs/modelling-the-domain.md
lvonk: When reconstructing from the event stream we call rubies `allocate()` method. This "constructs" an object without calling the initialiser.

See [AggregateRoot.load_from_history](https://github.com/zilverline/sequent/blob/bf75eb86c02a3dfc2884872b55760eefaae2a7e0/lib/sequent/core/aggregate_root.rb#L46)

docs/docs/modelling-the-domain.md
lvonk: Thanks for spotting this

mattgibson:
Hi. Thanks for a great framework! Sequent looks to be very useful for us. 

I've been starting to learn how to use it, but have found that some things in the docs are a little unclear, making it slow to get started. I've kept track of the questions that I've had in mind whilst going along, so perhaps it would be helpful when revising the docs for me to post them here:

- Is it better to have the attributes (e.g. the `title` of the `Post`) as fields on the command/event or to have a `PostAttributes` value object to encapsulate them so that the command/event has that as a single attribute instead? What are the pros and cons of each approach?
- How and when are the events applied when the aggregate is reconstructed from the event stream? The `initialize` method will be called each time a new object is instantiated, but there is logic in the examples that looks like it should only run when the initial `AddPost` command is handled.
- What does an `UpdatePost` command/event look like if I am using a HTML form to do the update (many fields)
- Why are `AggregateRoot`s and `Projector`s different classes? I know that this is the point of the CQRS pattern, but the docs and examples show no use case where the projectors do anything other than a 1-1 mapping of `Projector` to `AggregateRoot`. What would that look like?
- What is an `EventHandler`? Some examples have this, but the current docs don't. I see from the changelog that it's been renamed to `Projector`, but the examples were a source of confusion until I found this.
- What does the aggregate repository do? Does `add_aggregate` write to the DB? If so, which DB and what records are made? What does an `update_aggregate` operation look like?
- There are 3 DB tables in the `sequent_schema`: `event_records`, `command_records` and `stream_records`. What are they used for and when? Events I assume are used to rebuild the aggregates, but commands I don't see mention of being used and I don't know what stream are at all.
- Initially, it was not clear to me that the `aggregate_id` needed to be manually generated. I was assuming it would be automatically made on `repository.add_aggregate`
- Why is it not possible to do a zero downtime migration to add a new projection (leaving the existing ones alone)? Is this theoretically possible in future or will it never be?
- Is it OK to rename `aggregate_id` as `id` in the projections so as to be consistent with our other services? Does this matter or not?

Hopefully that's useful. I'm not very familiar with the CQRS pattern so perhaps some of these questions arise from seeing this project as my first experience with event sourcing. I'm happy to help with writing the documentation if that would be useful, but I don't know the answers to all of the above questions yet.

docs/docs/modelling-the-domain.md
mattgibson: Ah, I see. Never used this method before!

lvonk:
Hi @mattgibson thanks for your detailed feedback. We are continuously improving the docs and will certainly address your points.

lvonk:
Hi @mattgibson,

I have created a separate issue from your comments, since this PR will be merged.

See https://github.com/zilverline/sequent/issues/136

PR #139, opened 2018-09-24
Speed up offline replay by replaying recent events only

stephanvd:


spec/lib/sequent/migrations/view_schema_spec.rb
```diff
@@ -285,6 +286,26 @@ def self.version
         expect(MessageRecord.table_name).to eq 'message_records'
       end
 
+      context 'offline replaying with older events' do
+        it 'does not replay events older than 1 day' do
+          Timecop.freeze(1.week.ago)
```
lvonk: Make sure that in an after block `Timecop.return` is called

PR #154, opened 2018-11-23
Add Sequent::Secret as a Type

lvonk:
Secrets can be used to encrypt values in Commands so its value
is never stored in plain text in the command_records table
in the database.


docs/docs/concepts/types.md
```diff
@@ -8,14 +8,88 @@ types you are required to specify the type of an attribute.
...
+See [ValueObject](value-object.html) for a detailed explanation.
+
+## Lists
+
+A List of any of the types described here.
+
+## Sequent::Secret
+
+This is a special type designed to work with user input (HTML forms).
+You can use this type as follows:
```
stephanvd: Maybe mention that this is hashed using bcrypt and not reversible.

PR #165, opened 2019-02-14
BigDecimal.new is deprecated in ruby 2.6

sophiedeziel:
I'm using Sequent with ruby 2.6.x and I get some deprecation warnings:

```
warning: BigDecimal.new is deprecated; use BigDecimal() method instead.
```

I think I fixed all occurrences.


lvonk:
👍

PR #170, opened 2019-03-02
Event record attributes

BaylorRae:
First of all, I'd like to thank you for this framework. I'm just getting started with event sourcing and your project and documentation have been a great benefit for me to understand how this all works. Being familiar with ActiveRecord makes this project appealing and `ReplayOptimizedPostgresPersistor` looks awesome.

In my application I'm wanting to add [metadata] and [correlation/causation ids][correlation_causation] to the `event_records` table. I'm introducing this change to take advantage of `SerializesEvent#event=` to remove duplication and also allow a subclass of `EventRecord` to write additional columns from the event.

**Usage**

```ruby
class AppEventRecord < Sequent::Core::EventRecord
  def event=(event)
    super
    self.correlation_id = event.correlation_id if event.respond_to?(:correlation_id)
    self.causation_id = event.correlation_id if event.respond_to?(:causation_id)
    self.metadata_json = self.class.serialize_to_json(event.metadata) if event.respond_to?(:metadata)
  end
end
```

[metadata]: https://railseventstore.org/docs/request_metadata/
[correlation_causation]: https://railseventstore.org/docs/correlation_causation/

### DIFF
--- a/docs/docs/concepts/configuration.md
+++ b/docs/docs/concepts/configuration.md
@@ -105,6 +105,7 @@ For the most recent possibilities please check the `Sequent::Configuration` impl
 |event_record_class|The [class](event_store.html) mapped to the `event_records`|`Sequent::Core::EventRecord`|
 |stream_record_class|The [class](event_store.html) mapped to the `stream_records` table|`Sequent::Core::StreamRecord`|
 |snapshot_event_class|The event class marking something as a [Snapshot event](snapshotting.html)|`Sequent::Core::SnapshotEvent`|
+|event_record_hooks_class|The class with EventRecord life cycle hooks|`Sequent::Core::EventRecordHooks`|
 |transaction_provider|The transaction provider used by the [CommandService](command-service.html)|`Sequent::Core::Transactions::ActiveRecordTransactionProvider.new`|
 |event_publisher|The EventPublisher used by the [EventStore](event_store.html).|`Sequent::Core::EventPublisher.new`|
 |command_handlers|The list of [CommandHandlers](command-handler.html)|Empty|
--- a/lib/sequent/configuration.rb
+++ b/lib/sequent/configuration.rb
@@ -24,6 +24,8 @@ module Sequent
     DEFAULT_OFFLINE_REPLAY_PERSISTOR_CLASS = Sequent::Core::Persistors::ActiveRecordPersistor
     DEFAULT_ONLINE_REPLAY_PERSISTOR_CLASS = Sequent::Core::Persistors::ActiveRecordPersistor
 
+    DEFAULT_EVENT_RECORD_HOOKS_CLASS = Sequent::Core::EventRecordHooks
+
     attr_accessor :aggregate_repository
 
     attr_accessor :event_store,
@@ -33,6 +35,8 @@ module Sequent
                   :snapshot_event_class,
                   :transaction_provider,
                   :event_publisher
+
+    attr_accessor :event_record_hooks_class
 
     attr_accessor :command_handlers,
                   :command_filters
@@ -91,6 +95,8 @@ module Sequent
       self.event_store_schema_name = DEFAULT_EVENT_STORE_SCHEMA_NAME
       self.migrations_class_name = MIGRATIONS_CLASS_NAME
       self.number_of_replay_processes = DEFAULT_NUMBER_OF_REPLAY_PROCESSES
+
+      self.event_record_hooks_class = DEFAULT_EVENT_RECORD_HOOKS_CLASS
 
       self.offline_replay_persistor_class = DEFAULT_OFFLINE_REPLAY_PERSISTOR_CLASS
       self.online_replay_persistor_class = DEFAULT_ONLINE_REPLAY_PERSISTOR_CLASS
--- a/lib/sequent/core/event_record.rb
+++ b/lib/sequent/core/event_record.rb
@@ -4,6 +4,46 @@ require_relative 'sequent_oj'
 module Sequent
   module Core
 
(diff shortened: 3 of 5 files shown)


lvonk:
Hi @BaylorRae 

Thanks, this looks good. This is definitely something we want.

I am also thinking about extracting the `SerializesEvent` to be configurable. My feeling that that will be a bit "cleaner" approach, because there is no need to override the `event=` method in this case. What do you think?

Related: We are also about to add event_aggregate_id en event_sequence_number to the commands triggered by events in Workflows so you can create a full picture of why something happened in sequent. See the https://github.com/zilverline/sequent/tree/track-workflow-events branch

BaylorRae:
I like the idea of extracting `SerializesEvent` too and it would be much cleaner. As I'm looking at it though it seems like it _might_ work to have an `after_event_serialization` hook since the core attributes will/should always be assigned.

The upcoming features available in `track-workflow-events` will be a welcome change! This project is very cool.

```diff
diff --git a/lib/sequent/configuration.rb b/lib/sequent/configuration.rb
index 84b3c3a..64b02ce 100644
--- a/lib/sequent/configuration.rb
+++ b/lib/sequent/configuration.rb
@@ -24,6 +24,8 @@ module Sequent
     DEFAULT_OFFLINE_REPLAY_PERSISTOR_CLASS = Sequent::Core::Persistors::ActiveRecordPersistor
     DEFAULT_ONLINE_REPLAY_PERSISTOR_CLASS = Sequent::Core::Persistors::ActiveRecordPersistor
 
+    DEFAULT_AFTER_EVENT_SERIALIZATION_HOOK = -> (record, event) {}
+
     attr_accessor :aggregate_repository
 
     attr_accessor :event_store,
@@ -57,6 +59,8 @@ module Sequent
                 :versions_table_name,
                 :replayed_ids_table_name
 
+    attr_accessor :after_event_serialization
+
     def self.instance
       @instance ||= new
     end
@@ -96,6 +100,8 @@ module Sequent
       self.online_replay_persistor_class = DEFAULT_ONLINE_REPLAY_PERSISTOR_CLASS
       self.database_config_directory = DEFAULT_DATABASE_CONFIG_DIRECTORY
 
+      self.after_event_serialization = DEFAULT_AFTER_EVENT_SERIALIZATION_HOOK
+
       self.logger = Logger.new(STDOUT).tap {|l| l.level = Logger::INFO }
     end
 
diff --git a/lib/sequent/core/event_record.rb b/lib/sequent/core/event_record.rb
index 370346e..251ace1 100644
--- a/lib/sequent/core/event_record.rb
+++ b/lib/sequent/core/event_record.rb
@@ -17,6 +17,7 @@ module Sequent
         self.event_type = event.class.name
         self.created_at = event.created_at
         self.event_json = self.class.serialize_to_json(event)
+        Sequent.configuration.after_event_serialization.call(self, event)
       end
 
       module ClassMethods
diff --git a/spec/lib/sequent/core/event_record_spec.rb b/spec/lib/sequent/core/event_record_spec.rb
index dd6d904..25034de 100644
--- a/spec/lib/sequent/core/event_record_spec.rb
+++ b/spec/lib/sequent/core/event_record_spec.rb
@@ -86,6 +86,21 @@ describe Sequent::Core::EventRecord do
         record.event = event
         expect(record.organization_id).to eq('organization-id')
       end
+
+      it "invokes after_event_serialization" do
+        after_hook = spy(:after_event_serialization)
+        Sequent.configuration.after_event_serialization = after_hook
+
+        event = ExampleEvent.new(
+          aggregate_id: 'aggregate-id',
+          sequence_number: 1
+        )
+
+        record = ExampleRecord.new
+        record.event = event
+
+        expect(after_hook).to have_received(:call).with(record, event)
+      end
     end
   end
 end
```

lvonk:
Yes, an `after_event_serialization` hook seems good. I'd prefer a class method in a rather than a lambda in a configuration. So maybe something like this? Naming is up for improvement...

```ruby
class Configuration
  DEFAULT_AFTER_EVENT_SERIALIZATION_HOOK = SerializesEvent
end

class SerializesEvent
  def self.after_serialization(record, event)
    # noop
  end

  def event=(event)
    # ...
    Sequent.configuration.after_event_serialization.after_serialization(self, event)
  end
end

```

BaylorRae:
Defining the default hook as a class method is a great idea. This would even allow future event hooks to be wrapped in an `EventHooks` class. Which definitely needs a better name.

```ruby
class EventRecordHooks
  def self.before_serialization(record, event)
    # noop
  end

  def self.after_serialization(record, event)
    # noop
  end

  def self.before_save(record, event)
    # noop
  end

  # etc...
end
```

lvonk:
I think for now `EventRecordHooks` is probably best. If we come up with a better name, we can always rename later. Do you want to put this in this PR? Otherwise we will pick it up.

BaylorRae:
Awesome! I'll add it to this PR later today unless you want to add it now.

lvonk:
@stephanvd this looks good to me, can you take a look as well?

stephanvd:
Looks good to me! We can use this for `track-workflow-events` too.

I like the way this sets us up for more hooks in the future. Naming wise hook would work as callback is confusing with AR functionality. Or perhaps lifecycle?

lvonk:
@stephanvd Let's go with Hooks for now. Lifecycle, to me, implies creating, updating and deleting and such.

lvonk:
@BaylorRae can we merge this?

BaylorRae:
@lvonk yes, this is ready to be merged. Thanks for working through the naming and implementation on this.

PR #172, opened 2019-03-05
Add `after_commit` to workflows

### LINKED ISSUE #168: Design of the event handler
I am getting more and more comfortable in using Sequent and I think I came to the point where I understand what you mean by ["Sequent is an _opinionated_ CQRS and event sourcing framework"](https://www.sequent.io/). There are many ways to implement CQRS and opinions are needed to keep a framework lightweight without too much configuration and keep it's maintenance manageable.

There's one design choice that I try to work with right now and I'd like to discuss this with you. 

[Workflows](https://www.sequent.io/docs/concepts/workflow.html#)

> Workflows can be used to do other stuff (then updating a Projection) based on Events. Common tasks run by Workflows are:
>
> 1. Execute other Commands
> 2. Schedule something to run in the background
>
> In Sequent Workflows are committed in the same transaction as committing the Events.
>
> Since Workflows have nothing to do with Projections they do not run when doing a Migration.

First of all, I totally see how workflows and projections are both event handlers. When events happen, things happen. I also appreciate the elegance of everything being committed in the same transaction. In my understanding, a projection has no reason to fail and make the entire transaction to rollback. It should go smoothly. But my reasoning is not the same with workflows. The example provided, enqueuing a delayed job to send an email, has side effects. Which is also fine if it's the last step. Enqueuing a job is a very safe operation and is very unlikely to fail and rollback the transaction. 

There are small and bigger issues I'm trying to solve here. 

The first one is enqueuing a background job. It allows the transaction to be committed and a more "risky" task can be executed and the result will not impact previous events. A small annoyance is that often, my jobs will start executing even before the transaction has finished to commit. So sending a command from that background job using an `aggregate_id` that is not yet committed will fail. That's ok, I have retry mechanisms and the second try will succeed. But how to distinguish an aggregate that its command failed and one that is simply not ready yet?

Should there be some kind of event handler that can execute only if the events are actually committed? Let's say I'm sending an email after the UserCreated event and because of another command in the same queue fails. The database transaction will rollback but I can't take back the email. If it's ran in a background job, I can't cancel the job if it's already finished.

Should it be the `Sequent::Core::EventStore`'s responsibility to use a transaction in `store_events`  and publish events when it's actually committed? This is not a crazy idea if we embrace eventual consistency of the read side of the app. The `Sequent::Core::CommandService` could have it's own transaction provider that is not based on the database and could offer rollback features to workflows. But that will totally break the current behavior explained here: https://www.sequent.io/docs/concepts/command-service.html#

If I go ahead with my own refactor of `Sequent::Core::EventStore` and `Sequent::Core::CommandService`, would you be interested in a Pull Request or this is far from the direction you want to take with Sequent?

sophiedeziel:
As discussed in #168 , I created a helper to let a Workflow run side effects after a transaction commits. I also documented the feature.

It is basically used like this:

```ruby
class SendEmailWorkflow < Sequent::Workflow
  on UserCreated do |event|
    after_commit do
      # code with side effects
    end
  end
end
```

closes #168

### DIFF
--- a/spec/lib/sequent/core/workflow_spec.rb
+++ b/spec/lib/sequent/core/workflow_spec.rb
@@ -4,6 +4,8 @@ require 'sequent/test/event_handler_helpers'
 describe Sequent::Core::Workflow do
   include Sequent::Test::WorkflowHelpers
 
+  class CreateNotification < Sequent::Core::BaseCommand; end
+
   class UserWasRegistered < Sequent::Core::Event
     attrs email: String
   end
@@ -14,14 +16,24 @@ describe Sequent::Core::Workflow do
 
   class RegistrationWorkflow < Sequent::Core::Workflow
     on UserWasRegistered do |e|
-      execute_commands SendWelcomeEmail.new(email: e.email)
+      execute_commands CreateNotification.new(aggregate_id: e.aggregate_id)
+
+      after_commit do
+        execute_commands SendWelcomeEmail.new(email: e.email)
+      end
     end
   end
 
   let(:workflow) { RegistrationWorkflow.new }
 
+  let(:notification_command) { CreateNotification.new(aggregate_id: 'user') }
+  let(:email_command) { SendWelcomeEmail.new(aggregate_id: 'user', email: 'user@example.com') }
+
   it 'executes commands' do
-    when_event UserWasRegistered.new(aggregate_id: 'user', sequence_number: 1, email: 'user@example.com')
-    then_commands SendWelcomeEmail.new(aggregate_id: 'user', email: 'user@example.com')
+    fake_transaction_provider.transactional do
+      when_event UserWasRegistered.new(aggregate_id: 'user', sequence_number: 1, email: 'user@example.com')
+      then_commands notification_command
+    end
+    then_commands notification_command, email_command
   end
 end
--- a/docs/docs/concepts/workflow.md
+++ b/docs/docs/concepts/workflow.md
@@ -2,13 +2,13 @@
 title: Workflow
 ---
 
-Workflows can be used to do other stuff (then updating a Projection) based on [Events](event.html). Common
+Workflows can be used to do other stuff (than updating a Projection) based on [Events](event.html). Common
 tasks run by Workflows are:
 
 1. Execute other [Commands](command.html)
 2. Schedule something to run in the background
 
-In Sequent Workflows are committed in the same transaction as committing the Events.
+In Sequent, Workflows are committed in the same transaction as committing the Events.
 
 Since Workflows have nothing to do with Projections they do **not** run when doing a [Migration](migrations.html).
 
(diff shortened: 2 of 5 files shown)


lvonk:
Hi,

Thanks for this, looks great.

According to https://guides.rubyonrails.org/active_record_callbacks.html you need to handle any errors in the `after_commit` in order for **all** registered hooks to be called. For instance in a single transaction a user can register multiple hooks. If the first hook fails the second will not be executed by default. 
So I think it is nice if we at least document this. Perhaps we can an extra argument that will allow the users to ignore the error if they don't want to handle it `after_commit(ignore_errors: true) {}`

What do you think?

lvonk:
PS Thanks for improving the spelling errors!

spec/lib/sequent/core/workflow_spec.rb
```diff
@@ -14,14 +16,24 @@ class SendWelcomeEmail < Sequent::Core::BaseCommand
 
   class RegistrationWorkflow < Sequent::Core::Workflow
     on UserWasRegistered do |e|
-      execute_commands SendWelcomeEmail.new(email: e.email)
+      execute_commands CreateNotification.new(aggregate_id: e.aggregate_id)
+
+      after_commit do
+        execute_commands SendWelcomeEmail.new(email: e.email)
```
lvonk: The `execute_commands` will start a new transaction. Is that possible in an `after_commit` hook?

stephanvd: As long as it doesn't cause a loop, i.e. trigger another `UserWasRegistered`, this will be fine.

stephanvd:
Nice one @sophiedeziel. Love how simple the solution ends up being :)

sophiedeziel:
Thank you @lvonk for that information about handling errors in the `after_commit` hook. Yes, it is important to at least document this. `after_commit(ignore_errors: true) {}` is a very elegant solution. Should the default be `true` or `false`? There are good reasons for both options, but I think it should be `false` and make it clear that errors must be handled by the worker. It is, in my opinion, best to have loud errors than to have some hooks fail silently (unless you intentionally ignore them).

spec/lib/sequent/core/workflow_spec.rb
sophiedeziel: Yes, `execute_commands` will start a new transaction and yes, it is possible in an `after_commit`. I just tried it. @stephanvd is right but a loop can be created even without the `after_commit`, right?

lvonk: Yes indeed, so we do not have to do anything about that.

lvonk:
> Thank you @lvonk for that information about handling errors in the `after_commit` hook. Yes, it is important to at least document this. `after_commit(ignore_errors: true) {}` is a very elegant solution. Should the default be `true` or `false`? There are good reasons for both options, but I think it should be `false` and make it clear that errors must be handled by the worker. It is, in my opinion, best to have loud errors than to have some hooks fail silently (unless you intentionally ignore them).

Yes I think the default should be `false`

sophiedeziel:
I'm working on the `ignore_errors`'s implementation. There should be at least log entries about the errors. Do you have any recommendations on how to do it properly?

lvonk:
> I'm working on the `ignore_errors`'s implementation. There should be at least log entries about the errors. Do you have any recommendations on how to do it properly?

Maybe logging to `warn` (or `debug`, `info`) with the  `Sequent.logger` is sufficient for now

lvonk:
Thanks for all the work I am going to "squash and merge" this if that is okay with you.

sophiedeziel:
That's perfect! Thank you!

PR #174, opened 2019-03-15
Track workflow events

lvonk:
Workflows can execute new Commands based on Events in an async way. In order to have full
traceability of the entire Command - Event - Workflow - Command - Event chain the Command also
stores the event_aggregate_id and the event_sequence_number.


lib/sequent/core/workflow.rb
```diff
@@ -5,7 +5,31 @@ module Core
     class Workflow
       include Helpers::MessageHandler
 
+      def self.on(*message_classes, &block)
+        decorated_block = ->(event) do
+          begin
```
lvonk: `begin... end` is not needed in blocks.

lib/sequent/core/workflow.rb
```diff
@@ -5,7 +5,31 @@ module Core
...
+        super(*message_classes, &decorated_block)
+      end
+
+      def current_event
+        Thread.current[:handling_event]
+      end
+
       def execute_commands(*commands)
+        commands.each do |command|
+          if command.respond_to?(:event_aggregate_id) && command.event_aggregate_id.blank?
```
lvonk: Can we move this to the `Sequent.configuration.command_service.execute_commands`? This way it does not matter if the shortcut `execute_commands` or the real `execute_commands` is called.

spec/lib/sequent/core/command_service_spec.rb
```diff
@@ -1,14 +1,16 @@
 require 'spec_helper'
 
 class TestCommandHandler < Sequent::CommandHandler
-  class DummyCommand < Sequent::Core::Command; end
+  class DummyCommand < Sequent::Core::Command;
+  end
```
stephanvd: als deze op meerdere regels gaat mag de `;` weg. ben zelf wel van van single-line empty classes

spec/lib/sequent/core/command_service_spec.rb
```diff
@@ -1,14 +1,16 @@
...
 
   class DummyBaseCommand < Sequent::Core::BaseCommand
     attrs mandatory_string: String
     validates_presence_of :mandatory_string
   end
 
-  class NotHandledCommand < Sequent::Core::Command; end
+  class NotHandledCommand < Sequent::Core::Command;
+  end
```
stephanvd: `;`

PR #184, opened 2019-05-06
Add sql to changelog for the track workflow events feature

lvonk:


CHANGELOG.md
```diff
@@ -3,6 +3,16 @@
 Introduces optional `event_aggregate_id` and `event_sequence_number` columns to the `command_records` table.
 This enables keeping track of events causing commands in workflows.
 
+To add these columns to an existing event store you can use this sql to add them:
+
+Please not these sql statements use the `uuid` type for `aggregate_id`.
```
bforma: `Please note`

bforma: Can't we offer this in a migration file?

lvonk: For now we don't support migration files for the `sequent_schema`. Since the sequent_schema is pretty stable and tend to contain lot's of data I think it is better for now to let the users decide what their upgrade strategy is.

Issue #185, opened 2019-05-15
Should `Boolean` really be a class?


rintaun:
Currently, the `Boolean` constant used for validating `true`/`false` is a class:

https://github.com/zilverline/sequent/blob/0ef301d098735291fb3f9b723392d85baa8a249d/lib/sequent/core/ext/ext.rb#L32-L36

Is there a reason that this was made a class rather than a module? It doesn't seem particularly meaningful to _instantiate_ `Boolean`, and having it be a class is a hard conflict with a not-uncommon solution to a related problem:

```ruby
module Boolean; end
TrueClass.include Boolean
FalseClass.include Boolean
```

Can `Boolean` be changed to a module?

lvonk:
Hi,

The reason that it is a class is that it consistent with the other types. They all need to have a class method `deserialize_from_json`. This can also be done with a Module, but there wasn't any reason for that (maybe until now)

So what is the related problem you are referring to? Why do you want `TrueClass` and `FalseClass` to include `Boolean`?

rintaun:
Thanks for the response 😃 

> So what is the related problem you are referring to? Why do you want `TrueClass` and `FalseClass` to include `Boolean`?

The idea here is to allow `some_var.is_a? Boolean` to work if `some_var` is either `true` or `false`. I've seen it used in a number of places (including the codebase I am currently working on); one example is [this answer on Stack Overflow](https://stackoverflow.com/a/3028378).

PR #187, opened 2019-05-21
Make Boolean a module

rintaun:
See #185

In short, it seems like the goals of `Boolean` can be accomplished with it as a module, especially considering that there doesn't seem to be an expectation that `Boolean` will be instantiated. This also makes it more compatible with existing patterns, e.g.

```ruby
module Boolean; end
TrueClass.include Boolean
FalseClass.include Boolean

true.is_a? Boolean # => true
false.is_a? Boolean # => true
```

### DIFF
--- a/lib/sequent/core/ext/ext.rb
+++ b/lib/sequent/core/ext/ext.rb
@@ -29,7 +29,7 @@ class BigDecimal
   end
 end
 
-class Boolean
+module Boolean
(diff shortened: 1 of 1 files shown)


rintaun:
Tests are passing:

```
Finished in 34.57 seconds (files took 1.71 seconds to load)
364 examples, 0 failures, 1 pending
```

PR #192, opened 2019-06-18
Replace ActiveRecord::Base with Sequent::ApplicationRecord

rintaun:
Subclassing `ActiveRecord::Base` allows applications to isolate the connection for sequent, e.g. with a separate schema search path.


lvonk:
Thanks for this PR. Nice improvement.

Issue #190, opened 2019-06-04
Support patch versions for projections


lvonk:
Currently there is a single strategy of building projections:

The new projection is created in a background process by replaying all events. At the end the application needs a little downtime to get the events added after the new projection was created.

This strategy is fast and works great if the projections don't need many events (let's say less then 5.000.000) to rebuild its state. Also having a single strategy is simple and easy to understand.

During a lot of migrations we are just adding data to the new projection. Data that previously did not exist. In this case it makes no sense to rebuild the entire projection again. Although downtime is still minimized rebuilding a projection in the background can take significant amount of time. For example an event store of > 200M events and replaying ~ 15M events can take up to an hour in the background.

So what we like to have is a way of allowing new data in projections (adding columns to a table) without needing to rebuild the entire projection. This can be achieved by adding a patch version. 

Proposal to implement this:

```ruby
class ViewSchemaMigrations < Sequent::Migrations::Projectors
  def self.version
    '1.1'
  end
  def self.versions
    {
      '1' => [InvoiceProjector, EstimateProjector],
      '1.1' => [InvoiceProjector],
    }
  end
end
```

We then also need the following sql files:

`invoice.sql` containing the initial table creation.
`invoice_1.1.sql` containing the addition of the columns.

The `Sequent::Migrations::ViewSchema::Versions` needs to have an extra column `patch_version` to store the current patch version (default 0).

In `Sequent::Migrations::ViewSchema.migrate_offline` we can implement applying the patch version.

We also need to think about how to deal with rollbacks and if we want to support the following scenario: upgrading from version '1.0' via '1.1' directly to version '2.0'.

cc @stephanvd @bforma @mvandiepen

stephanvd:
I like this solution! Nice that it's opt-in like that. One doesn't need to know about it until they're at scale and looking for an advanced topic like this.

Patches don't really need projectors. To be more explicit about what's happing in patch versions so we can point to the table to alter. i.e.:

```
  def self.versions
    {
      '1' => [InvoiceProjector, EstimateProjector],
      '1.1' => [AlterTable.new('invoice_records')],
    }
  end
```

> We also need to think about how to deal with rollbacks and if we want to support the following scenario: upgrading from version '1.0' via '1.1' directly to version '2.0'.

**Upgrading 1.0 -> 2.0:**

[...]

PR #194, opened 2019-07-02
Introduce alter table without replaying

lvonk:
See #190


lvonk:
Ping @stephanvd .

TODO:
- documentation

lib/sequent/migrations/planner.rb
```diff
@@ -104,7 +104,6 @@ def map_to_migrations(migrations)
               fail "Missing file #{alter_table_sql_file_name} to apply for version #{version}" unless File.exist?(alter_table_sql_file_name)
               migration.copy(version)
             else
-              migration.is_a?(Sequent::Projector)
```
lvonk: Better to check in `elsif` and fail if it is something else.

Issue #193, opened 2019-07-02
Strict apply values for events


lvonk:
Sequent currently ignores `apply ThisHappened, does_not_exist: 'value'` attributes that are
not declared on an Event. The same applies if you put a value in that is not of the correct type. Type checking and casting is done in the `CommandService`, but this leaves room for programming errors.

Since Events are forever it is better to be more strict about this so no accidental errors occur. `apply` should fail if the attributes do not exist, or the types are wrong.

PR #195, opened 2019-07-02
Ensure correct attrs

stephanvd:
Right now unspecified attributes will be silently ignored. This PR raises an error if an unknown attr is passed in.

For example when we have some event:

```rb
class SomeEvent < Sequent::Core::Event
  attrs message: String
end
```

We instantiate it with unknown attributes:
```rb
SomeEvent.new(message: 'hello', something: 'this should raise', something_else: 'and this')
```

It gives us a nice descriptive error:
```rb
Sequent::Core::Helpers::AttributeSupport::UnknownAttributeError:
  SomeEvent does not specify attrs: something, something_else
```

This works on anything that includes `AttributeSupport` like `Command` and `ValueObject`.

Implements #193

### DIFF
--- a/lib/sequent/core/command.rb
+++ b/lib/sequent/core/command.rb
@@ -19,7 +19,7 @@ module Sequent
       include ActiveModel::Validations::Callbacks
       include Sequent::Core::Helpers::TypeConversionSupport
 
-      attrs created_at: DateTime
+      attrs aggregate_id: String, created_at: DateTime
 
       def initialize(args = {})
         update_all_attributes args
--- a/lib/sequent/core/command.rb
+++ b/lib/sequent/core/command.rb
@@ -19,7 +19,7 @@ module Sequent
       include ActiveModel::Validations::Callbacks
       include Sequent::Core::Helpers::TypeConversionSupport
 
-      attrs aggregate_id: String, created_at: DateTime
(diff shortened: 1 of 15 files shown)


lib/sequent/core/helpers/attribute_support.rb
```diff
@@ -60,6 +62,7 @@ def attrs(args)
             class_eval <<EOS
               def update_all_attributes(attrs)
                 super if defined?(super)
+                ensure_known_attributes(attrs)
```
lvonk: This may only throw an error on instatiation, not deserialization. Otherwise existing events can not be deserialized.

lib/sequent/core/command.rb
```diff
@@ -19,7 +19,7 @@ class BaseCommand
       include ActiveModel::Validations::Callbacks
       include Sequent::Core::Helpers::TypeConversionSupport
 
-      attrs created_at: DateTime
+      attrs aggregate_id: String, created_at: DateTime
```
lvonk: `Command` has the `aggregate_id`, not `BaseCommand`

stephanvd:
[Sequent: Validate event attributes](https://trello.com/c/bNCBP1Mm/3-sequent-validate-event-attributes)

PR #199, opened 2019-07-16
End Modelling the Domain with passing tests

reentim:
I'm just learning Sequent, so this may not make any sense! :)

Towards the end of this chapter, the reader has implemented behaviour for the _it ignores case in usernames_ example, and all tests pass. But then the reader does some refactoring, and unless they also update `post_command_handler_spec.rb` (is this a deliberate exercise?), they'll finish the chapter with failing tests.

I've updated the guide to first update the test, and then do the refactoring, guideded by the tests. This is how I'd do it, and seems somewhat in keeping with the guide's earlier test-driven approach.

One issue doing it this way, is that the tests all pass _before_ the refactoring is complete, which was a surprise.

Given these expectations:

```ruby
then_events(
  # ...
  PostAuthorChanged.new(
    aggregate_id: aggregate_id,
    author_aggregate_id: author_aggregate_id,
    sequence_number: 2,
  ),
  # ...
)
```

and these events and event handler, not updated to use `author_aggregate_id`:

```ruby
class PostAuthorChanged < Sequent::Event
  attrs author: String
end

# ...

on PostAuthorChanged do |event|
  @author = event.author
end
```

I would have expected some error, but it passes.

What changes could I make to the test, to force the changes to `author_aggregate_id` above?

### DIFF
--- a/docs/docs/modelling-the-domain.md
+++ b/docs/docs/modelling-the-domain.md
@@ -492,9 +492,35 @@ class Usernames < Sequent::AggregateRoot
 end
 ```
 
-The last thing we need to do is refactor the `PostCommandHandler` to use not the Author name, but `aggregate_id` of the `Author`.
+The last thing we need to do is refactor out `Author` name, and instead use the `Author` `aggregate_id`.
 
-In `lib/post/commands.rb` change
+In `spec/lib/post/post_command_handler_spec.rb`:
+
+```ruby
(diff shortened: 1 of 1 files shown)


lvonk:
Wow, thanks a lot for this. I think that https://github.com/zilverline/sequent/pull/195 was the reason the test did not fail. The latest sequent version fails if you do not specify the correct attrs.

Could you try to run this against the latest master branch?

PR #214, opened 2020-01-03
Add dry run option

lvonk:
When using dry run the given commands and resulting events are not stored
but recorded for inspection purposes.
It will also record which Projectors and Workflows would have been invoked
as a result of the Events applied.
Due to the fact that the Sequent.configuration is a singleton and shared
among Threads dry run is not Thread safe.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ You can start the documentation locally via:
 ```
 cd docs
 bundle install
-bundle exec jekyll serve
+bundle exec jekyll serve --livereload
 ```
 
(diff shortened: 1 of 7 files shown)


lvonk:
TODO: Find a better name for 'Advanded' in the documentation. Maybe just make it separate chapters.

lvonk:
> TODO: Find a better name for 'Advanded' in the documentation. Maybe just make it separate chapters.

For now stick to 'advanced.'

PR #228, opened 2020-02-23
Avoid using RSpec globals

nbibler:
The general suggestion with RSpec is to configure it with `config.disable_monkey_patching!` which means the `describe` method isn't in the global namespace. This works with or without that setting enabled.


lvonk:
Thanks, but I will keep this as is:

I like the global `describe` over `Rspec.describe` because I think it is more readable (and less typing). So in this case I think it is okay to have an occasional monkey patch (in a spec file).

PR #234, opened 2020-06-22
Support Hash type for attrs

bforma:


lvonk:
Nice, but I don't this will work with a Hash containing ValueObjects... can you check? Because that might be confusing.

bforma:
> Nice, but I don't this will work with a Hash containing ValueObjects... can you check? Because that might be confusing.

Sure, I'll add some specs with 'complex' Hash structures.

lvonk:
ping @bforma

lvonk:
Closing due to inactivity

spec/lib/sequent/core/helpers/type_conversion_support_spec.rb
```diff
@@ -264,6 +264,47 @@ class CommandWithDateTime < Sequent::Core::BaseCommand
 
   end
 
+  context Hash do
+    class CommandWithHash < Sequent::Core::BaseCommand
+      attrs value: Hash
```
lvonk, line 269: I don't think a Hash containing a ValueObject works. Can you add a test for this? I think if it does not work we should fail fast, and document that it is not supported.

lvonk (commented):
Can you check if a Hash containing a ValueObject works?

spec/lib/sequent/core/helpers/type_conversion_support_spec.rb
bforma: Yes I was going to fix this.

PR #242, opened 2020-09-11
Quick update on Modelling your Domain Docs 

victorfeijo:
I believe the docs is trying to say that the "Next thing we want to do is create a `AuthorCommandHandler`" instead of a `UserCommandHandler`.  It was probably confused with the `Usernames` instance just mentioned on the same paragraph. Please let me know if this fix is correct.


lvonk (approved):

lvonk:
Thank you!

Issue #246, opened 2020-09-25
commands scheduled in after_commit workflow blocks are not executed


TreyE:
Here's my current project: https://github.com/TreyE/entity_journal

I'm having an issue with execution of any events in an after_commit callback in my workflows.

For a good example, see lines 5-11 of https://github.com/TreyE/entity_journal/blob/master/app/sequent/entity_update_notifications/workflows.rb.

If I remove the "after_commit" block, my secondary command is fired with no issue.

Some test code that will let you see exactly what is happening by triggering the exact code path is:

`Sequent.command_service.execute_commands ConsumeEntityUpdateNotification.new(aggregate_id: "abcde", event_body: "some xml")`

If it also helps, I'm running the default rails setup with Puma on Ruby 2.6.6.

Update:  I'm running Postgres 12.  Wondering if it's related to: #216 ?

Update part 2 - looks like it never enters: https://github.com/zilverline/sequent/blob/master/lib/sequent/core/command_service.rb#L49 during the run of the after commit commands.

Update part 3 - what it comes down to is it looks like Sequent::Util.skip_if_already_processing is not re-entrant during recursive calls, which is what 'execute_service' in an after commit block does, because of the way the ActiveRecord transaction provider behaves with after_commit hooks.

I've got a simple solution with wraps it in a thread, dodging the thread local lock, but I don't like it.  I'm seeing if I can do something else.

lvonk:
Hi nice to see you are trying out Sequent! And yes this indeed seems to be a bug, good catch.

Looking at your code I do wonder why you chose to use `after_commit` to schedule new commands though. If you are going to do this synchronously you might as well leave the `after_commit` out. This way you don't have to deal with possible weird state in which the real transaction commits, but the `after_commit` fails.
The initial intent of `after_commit` was to be able to schedule background jobs in non postgres backed background workers like Sidekiq.

TreyE:
I can understand the intention but if so it's not well documented that it's not supported.

In our case we are using sequent in an event sourcing system to process large amounts of highly unreliable data.  In this case it is entirely possible we will get one event in and the transaction will fail.  Under certain conditions, I want to keep the progress in the transactions that has been successful until the failure.

[...]

PR #247, opened 2020-09-27
Correctly execute commands in workflow after-commit blocks.

TreyE:
Addresses the issue with #246.

### DIFF
--- a/lib/sequent/core/command_service.rb
+++ b/lib/sequent/core/command_service.rb
@@ -43,10 +43,22 @@ module Sequent
         event_store.remove_event_handler(clazz)
       end
 
+      def reset_lock_key
+        Thread.current[:sequent_command_service_lock_key] = :command_service_process_commands
+      end
+
+      def set_after_commit_lock
+        Thread.current[:sequent_command_service_lock_key] = :sequent_command_service_after_commit_lock_key
+      end
+
       private
 
+      def lock_key
+        Thread.current[:sequent_command_service_lock_key] ||= :command_service_process_commands
+      end
+
       def process_commands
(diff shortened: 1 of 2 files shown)


lib/sequent/core/command_service.rb
```diff
@@ -43,10 +43,22 @@ def remove_event_handler(clazz)
...
+
       private
 
+      def lock_key
+        Thread.current[:sequent_command_service_lock_key] ||= :command_service_process_commands
+      end
+
       def process_commands
-        Sequent::Util.skip_if_already_processing(:command_service_process_commands) do
+        Sequent::Util.skip_if_already_processing(lock_key) do
```
lvonk, line 61: Can't we better just "unset" the `command_service_process_commands` after the while block? This way we do not have to introduce new Thread local keys.

So `process_commands` will be:
```ruby
Sequent::Util.skip_if_already_processing(:command_service_process_commands) do
  begin                                                                       
    transaction_provider.transactional do                                     
      while(!command_queue.empty?) do                                         
        process_command(command_queue.pop)                                    
      end
      Sequent::Util.done_processing(:command_service_process_commands)                                            
    end                                                                       
  ensure                                                                      
    command_queue.clear                                                       
    repository.clear                                                          
  end                                                                         
end
```

lvonk (commented):
Thanks for providing a fix 👍 . I think unsetting the "processing" key will also work and keeps the code  clearer imo because we have less (different) thread  locals.
Also it would be really nice if we could have a test that proves this bug and that the provided fix works :-). If you don't have time, or need help with this let me know.

lvonk:
This issue is addressed in https://github.com/zilverline/sequent/pull/249 @TreyE

PR #249, opened 2020-10-02
Fix re-entry of process_commands in after_commit hook

### LINKED ISSUE #246: commands scheduled in after_commit workflow blocks are not executed

lvonk:
Reset the re-entry lock after command queue is empty
so when the after_commit queue executes another
command the process command is re-entrant again.
Also make sure that after_commit is a queue and
and the registered blocks are popped otherwise
they will be executed on each new commit.

Fixes #246

### DIFF
--- a/lib/sequent/core/command_service.rb
+++ b/lib/sequent/core/command_service.rb
@@ -52,6 +52,7 @@ module Sequent
               while(!command_queue.empty?) do
                 process_command(command_queue.pop)
               end
+              Sequent::Util.done_processing(:command_service_process_commands)
             end
           ensure
             command_queue.clear
--- a/lib/sequent/core/transactions/active_record_transaction_provider.rb
+++ b/lib/sequent/core/transactions/active_record_transaction_provider.rb
@@ -7,7 +7,9 @@ module Sequent
           Sequent::ApplicationRecord.transaction(requires_new: true) do
             yield
           end
-          after_commit_queue.each &:call
+          while(!after_commit_queue.empty?) do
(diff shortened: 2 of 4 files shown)


lvonk:
Hi @TreyE can you check if this fix works for you?

stephanvd (approved):

Issue #251, opened 2021-01-25
Add Ruby 3 / Rails 6 support


JulienItard:
Hello, just want to know if it's planned to add support for ruby 3 and rails 6 ? thanks :)

lvonk:
Hi,

Yes this is definitely something we are going to support. Rails 6 is already supported btw. We don't know a precise date yet though.

lvonk:
Hi @JulienItard you can already take a look at https://github.com/zilverline/sequent/pull/252. We are going to do some extensive testing with our own apps to make sure it works. After that we will merge and release a new version of the gem.

lvonk:
Sequent 4.0 supports Ruby 3.

PR #252, opened 2021-01-27
ruby 3

lvonk:
- Upgrade to ruby 3.0
- Let ~~travis~~ github actions run with ruby 3.0


lvonk:
Fixes https://github.com/zilverline/sequent/issues/251

Locally one spec (always the first) fails from the `spec/lib/sequent/core/persistors/active_record_persistor_spec.rb` spec. The first time it tries to drop the in the test created database and fails because it is accessed by another user. The 2nd and 3rd test of the file can drop the created database.
Can not figure out yet why, let's see if the problem is the same on travis.

lvonk:
Spec failed due to https://github.com/rails/rails/issues/40029, applied the work around and will continue to upgrade to latest AR. We should not use AR 6.0.3 I guess.

.ruby-version
```diff
@@ -1 +1 @@
-2.7.1
+3.0.0
```
bforma, line 1: 🚀

bforma (approved):
Looks good! 💪

PR #262, opened 2021-07-08
Support 6.1.4

stephanvd:


spec/lib/sequent/core/command_service_spec.rb
```diff
@@ -96,8 +96,8 @@ def password
       )
     end
 
-    context 'and dutch as error locale' do
-      before { Sequent.configuration.error_locale_resolver = -> { :nl } }
+    context 'and dutch as the locale' do
+      before { I18n.locale = :nl }
```
lvonk, line 100: I think this is also to test the `Sequent.configuration.error_locale_resolver` ?

lvonk (changes requested):
Question about the changing of the `Sequent.configuration.error_locale_resolver` in the spec?

spec/lib/sequent/core/command_service_spec.rb
stephanvd: Why is this different from I18n.locale?

lvonk: This is for translating **just the command validation errors** in the correct local. So other things will still use the default translation. We added this to return errors in english in an api for instance, but still be able to use another language as language for the normal flow.

lvonk: Why did it fail?

stephanvd: nvm, fixed it :)

stephanvd: Apparently I18n.with_locale now doesn't cache the validation messages, so we had to move the block to where it's called.

lib/sequent/core/command_service.rb
```diff
@@ -68,9 +68,7 @@ def process_command(command)
 
         filters.each { |filter| filter.execute(command) }
 
-        I18n.with_locale(Sequent.configuration.error_locale_resolver.call) do
-          raise CommandNotValid.new(command) unless command.valid?
-        end
+        raise CommandNotValid.new(command) unless command.valid?
```
lvonk, line 71: Weird that this worked before than...

lvonk (approved):
If tested against our own app we can merge.

lib/sequent/core/command_service.rb
stephanvd: > Apparently I18n.with_locale now doesn't cache the validation messages, so we had to move the block to where it's called.

lib/sequent/core/event_store.rb
```diff
@@ -62,7 +62,7 @@ def load_events_for_aggregates(aggregate_ids)
         streams = Sequent.configuration.stream_record_class.where(aggregate_id: aggregate_ids)
 
         query = aggregate_ids.uniq.map { |aggregate_id| aggregate_query(aggregate_id) }.join(" UNION ALL ")
-        events = Sequent.configuration.event_record_class.connection.select_all(query).map! do |event_hash|
+        events = Sequent.configuration.event_record_class.connection.select_all(query).map do |event_hash|
```
stephanvd, line 65: @lvonk Do you know why this was a `.map!`, which is deprecated now. The only reason I can think of is a slight memory optimization, but am I missing anything?

lvonk: Yes I think so too

stephanvd:
Initial testing of the new migrator encoding with our own app in development shows the desired results, can you take another look @lvonk ?

lib/sequent/support/database.rb
```diff
@@ -46,7 +46,7 @@ def self.execute_sql(sql)
 
       def self.create_schema(schema)
         sql = "CREATE SCHEMA IF NOT EXISTS #{schema}"
-        if user = ActiveRecord::Base.connection_config[:username]
+        if user = ActiveRecord::Base.connection_db_config.configuration_hash[:username]
```
lvonk, line 49: Is this backward compatible to earlier versions of AR?

lib/sequent/core/persistors/replay_optimized_postgres_persistor.rb
```diff
@@ -346,7 +346,10 @@ def clear
         private
 
         def cast_value_to_column_type(clazz, column_name, record)
-          Sequent::ApplicationRecord.connection.type_cast(record[column_name.to_sym], @column_cache[clazz.name][column_name])
+          data = clazz.arel_table.type_cast_for_database(column_name, record[column_name.to_sym])
+          return data unless data.respond_to?(:encoder)
```
lvonk, line 350: Is there a case when this is not the case?

lvonk (approved):
Nice!

lvonk:
We should add a build to our build matrix as well right? Currently `6.1.4` is not tested in our build.

lib/sequent/core/persistors/replay_optimized_postgres_persistor.rb
stephanvd: For strings it returns the string right away instead of the encoder.

lib/sequent/support/database.rb
stephanvd: Good catch, added back support for the legacy config

lvonk (approved):

PR #264, opened 2021-07-19
FIX: enable snapshot setting errors

HEROGWP:
- Run rake task `sequent:snapshots:set_snapshot_threshold` will get `args` missing
- Run ` Sequent::Core::SnapshotCommand` command will get event_store undefined method error in AggregateSnapshotter

### DIFF
--- a/lib/sequent/core/aggregate_snapshotter.rb
+++ b/lib/sequent/core/aggregate_snapshotter.rb
@@ -16,7 +16,7 @@ module Sequent
     class AggregateSnapshotter < BaseCommandHandler
 
       on SnapshotCommand do |command|
-        aggregate_ids = repository.event_store.aggregates_that_need_snapshots(@last_aggregate_id, command.limit)
+        aggregate_ids = Sequent.configuration.event_store.aggregates_that_need_snapshots(@last_aggregate_id, command.limit)
         aggregate_ids.each do |aggregate_id|
           take_snapshot!(aggregate_id)
         end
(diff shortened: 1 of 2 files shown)


lvonk (approved):
Thanks for fixing 👍🏼

lib/sequent/core/aggregate_snapshotter.rb
```diff
@@ -16,7 +16,7 @@ class TakeSnapshot < Sequent::Core::Command
     class AggregateSnapshotter < BaseCommandHandler
 
       on SnapshotCommand do |command|
-        aggregate_ids = repository.event_store.aggregates_that_need_snapshots(@last_aggregate_id, command.limit)
+        aggregate_ids = Sequent.configuration.event_store.aggregates_that_need_snapshots(@last_aggregate_id, command.limit)
```
lvonk, line 19: I think this only raises an error if you did not setup `AggregateSnapshotter` in your sequent initializer as a `CommandHandler` like so:

```ruby
Sequent.configure do |config|
  config.command_handlers = [
    Sequent::Core::AggregateSnapshotter,
  ]
end
```

Besides that, this is still a good improvement to reference the event store using the `Sequent.configuration`

HEROGWP: In fact, I did the same settings as you, and the error occurred when I executed this command

```ruby
Sequent.command_service.execute_commands Sequent::Core::SnapshotCommand.new(limit: 10) 
```

lvonk (approved):

PR #266, opened 2021-07-28
Improve ReplayOptimizedPostgresPersistor::Index

lvonk:
Two improvements:

1. The order of the columns in the indexes no longer matters
2. The type of the column in the where and index
     (String or Symbol) no longer matters. Internally everything
    is stored as String.

For now on if you define an Index like:

```
{FooRecord => [[:aggregate_id, :tenant_id]]}
```

All queries below will hit the Index:

```
{aggregate_id: 1, tenant_id: 2}
{'aggregate_id' => 1, 'tenant_id' => 2}
{'tenant_id' => 2, 'aggregate_id' => 1}
```


mvandiepen (approved):
Looks good 👍

PR #267, opened 2021-08-13
Allow database config to be set using a postgres connection URL

berkes:
I prefer to use a `DATABASE_URL` env var to set my database config. Many tooling and hosting environments default to such an URL, including HEROKU and AWS-RDS.

The database config can not always deal with it. Several methods, like `drop!` would flunk as they expect the config to have keys like `database` or `username` and such. 

This patch leverages ActiveRecords API to convert any valid Activerecord config hash into a normalized config hash including keys such as `database` or `username`. Which then can be used anywhere we expect config with certain keys.

It does create a HashWithIndifferentAccess, so normalizing the keys to symbols makes it behave a tad more like one expects (and not accidentally creating a key `"database"` next to a key `:database`), so this patch made that apparent by changing the keys to symbols.

I did not check on older AR versions, AFAIK all supported AR versions have and had this API to normalize the config hash.
I did not run a linter, as I could not find any configured in this project. Please tell me if I violate syntax formatting conventions.

A spec is added to test that a config hash with only an URL can connect to the test database.

### DIFF
--- a/lib/sequent/support/database.rb
+++ b/lib/sequent/support/database.rb
@@ -19,17 +19,18 @@ module Sequent
         fail ArgumentError.new("env is mandatory") unless env
 
         database_yml = File.join(Sequent.configuration.database_config_directory, 'database.yml')
-        YAML.load(ERB.new(File.read(database_yml)).result)[env]
+        config = YAML.load(ERB.new(File.read(database_yml)).result)[env]
+        ActiveRecord::Base.resolve_config_for_connection(config)
       end
 
       def self.create!(db_config)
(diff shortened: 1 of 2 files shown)


lvonk:
TIL `resolve_config_for_connection` 👍🏼 . Let's see how I can run our github actions matrix on Pr's from forks without merging it into master first.... 🤔

lvonk:
Hi @berkes I made some changes to the github actions configuration so it also runs the builds for pr's from forked repos. Would you mind rebasing your PR on the current master and push again so it will trigger the build. Then we know for sure it works on all the supported versions.

berkes:
Bit late to the party, but I was travelling. Sorry for not responding. And thanks a million for your time!

PR #276, opened 2021-09-06
Add rubocop

lvonk:


lvonk:
It seems that reviewdog does not use Sequent's `.rubocop.yml`

lvonk:
> It seems that reviewdog does not use Sequent's `.rubocop.yml`

Yes it is, it's only displaying the findings in the pr.

PR #277, opened 2021-09-18
:bug: private method `resolve_config_for_connection' called

haoxilu:
> - The `resolve_config_for_connection` has been moved to private method in the recent changes
> - And its logic has been changed as well. 

- NoMethodError: private method `resolve_config_for_connection'
called for ActiveRecord::Base:Class

image

### DIFF
--- a/.github/workflows/rspec.yml
+++ b/.github/workflows/rspec.yml
@@ -21,7 +21,7 @@ jobs:
       matrix:
         include:
           - ruby-version: 3.0.0
-            active-star-version: '< 6.1.5'
+            active-star-version: '< 7'
           - ruby-version: 3.0.0
             active-star-version: '6.1.4'
           - ruby-version: 3.0.0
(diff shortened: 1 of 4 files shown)


haoxilu:
WIP

lvonk:
Hi @haoxilu thanks for your PR. I don't see why we need to upgrade the supported ActiveRecord versions though? We have a rather rigorous testing of our own applications before we do that. So I'd opt for removing that for this PR and keeping the AR versions the same.

haoxilu:
> Hi @haoxilu thanks for your PR. I don't see why we need to upgrade the supported ActiveRecord versions though? We have a rather rigorous testing of our own applications before we do that. So I'd opt for removing that for this PR and keeping the AR versions the same.

okk. I see. But I need to support the newest version of ActiveRecord to my own app. So Please close the PR if you don't need that.

Issue #278, opened 2021-09-21
handles_message? is not matching the given class


haoxilu:
<img width="722" alt="image" src="https://user-images.githubusercontent.com/7600503/134097482-7a576a22-3eeb-4b30-a709-7247a5da640d.png">

<img width="851" alt="image" src="https://user-images.githubusercontent.com/7600503/134097466-2fd3cbc9-4a91-4160-ab91-366b2a8c72de.png">

lvonk:
Not sure why this happens for you, perhaps byebug is reloading class definitions. This is not a bug in sequent as far as i can tell.

PR #279, opened 2021-09-21
:recycle: Using class.name as hash key in Message Handler

haoxilu:
https://github.com/zilverline/sequent/issues/278

Fix the issue handles_message? is not matching the given class

image

image


lvonk:
Not sure why byebug gives this error for you, perhaps it reloaded your class definition? Anyway, this is not a bug as far as i can tell.

PR #280, opened 2021-09-24
Implement command middleware

bforma:
Usage:

```ruby
class MyCommandMiddleware
  def call(command)
    # ...
    yield 
  end
end

Sequent.configure do |config|
  config.command_middleware.add(MyCommandMiddleware.new)
end
```

The middleware chain is invoked for each command passed to `Sequent.configuration.command_service.execute_commands(...)`.


lvonk:
I like the use of Middleware of the filters we currently use. This is also more inline with how e.g. Rack works with Middleware. We should add documentation for this and probably (later) deprecate command_filters in favor of this.

bforma:
> I like the use of Middleware of the filters we currently use. This is also more inline with how e.g. Rack works with Middleware. We should add documentation for this and probably (later) deprecate command_filters in favor of this.

Yes, and we can also add some default middleware.

bforma:
Any word on this? We could use this.

lvonk:
Yes, we do need some documentation for this, that would be really helpful 🙏🏼 . Is that something you can add?

bforma:
Sure.

bforma:
Can this be merged now?

PR #284, opened 2021-11-25
Implement attribute upcasting

bforma:


lvonk (approved):
Looks good. We should also add documentation for this.

PR #285, opened 2021-11-30
Fix private method call to resolve_config_for_connection

lvonk:
See https://github.com/zilverline/sequent/pull/277.


lvonk:
Thanks to @haoxilu for the initial PR.

bforma (approved):

PR #298, opened 2022-01-05
Bug/fix dry run nested workflows

ed1001:


lib/sequent/util/dry_run.rb
```diff
@@ -186,8 +187,19 @@ def self.these_commands(commands)
                         current_event_store
                       end
 
+        # Set event handlers to classes not instances for dry run. This allows for
+        # variation of class constructor signatures.
+        #
+        # If in test mode use configuration, else set to all classes that inherit from
+        # handler parent classes to encompass possibly nested handlers.
+        event_handler_classes = if ENV['RACK_ENV'] == 'test'
```
lvonk, line 195: Why not always use `[*Sequent::Core::Workflows.all, *Sequent::Core::Projectors.all]`?

lvonk (changes requested):

lib/sequent/util/dry_run.rb
ed1001: Then the dry run spec would pick up workflows from previous specs

lvonk: We should fix this in the spec (perhaps by stubbing this in the dry_run spec) rather than implementing this in code. Also, not everyone uses `RACK_ENV`, we should get rid of that.

lvonk (approved):
:shipit:

PR #300, opened 2022-01-05
bump AR version

gen1321:
Slightly bump AR version


lvonk:
Could you also update the https://github.com/zilverline/sequent/blob/master/.github/rspec.yml so the code is tested against the latest `6.1.4` ?

gen1321:
Sure

lvonk:
Thanks!

Issue #293, opened 2021-12-21
Printed location of the app after `sequent new myapp` is pointing to wrong dir


lvonk:
Run

```bash
gem install sequent
sequent new myapp
```

It prints:

```
Success!

Your brand spanking new sequent app is waiting for you in:
  /Users/foobar/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/sequent-4.1.0/bin/myapp
```

Which is not the location of the `myapp`

lvonk:
Fixed by https://github.com/zilverline/sequent/pull/304

PR #304, opened 2022-01-30
Print the current working dir instead of location of gem

lvonk:


lvonk:
Fixes https://github.com/zilverline/sequent/issues/293

Issue #294, opened 2021-12-21
File `.ruby-version` is not copied from the template project


lvonk:
When generating a new app with sequent 4.1.0 the `.ruby-version` file is missing from the template dir
This results in falling back to the default version which, in case it is not 3.0.0, in the following error:

```
Your Ruby version is 2.7.2, but your Gemfile specified 3.0.0
```

lvonk:
See https://github.com/zilverline/sequent/issues/288

Issue #295, opened 2021-12-21
Printed message after generating app misses `bundle install`


lvonk:
After generating a `newapp` it says:

```
To finish setting up your app:
  cd invoicing
  bundle exec rake sequent:db:create
  bundle exec rake sequent:db:create_view_schema
  bundle exec rake sequent:migrate:online
  bundle exec rake sequent:migrate:offline
```

This should be
```
To finish setting up your app:
  cd invoicing
  bundle install
  bundle exec rake sequent:db:create
  bundle exec rake sequent:db:create_view_schema
  bundle exec rake sequent:migrate:online
  bundle exec rake sequent:migrate:offline
```

PR #305, opened 2022-01-30
Fix setup sequent and spec testing setup

lvonk:


lvonk:
Fixes https://github.com/zilverline/sequent/issues/294 and https://github.com/zilverline/sequent/issues/295

PR #307, opened 2022-02-08
Add load_until to load_events_for_aggregates

emielhagen:
In order to load aggregates upon a certain point in time this PR introduces the load_until argument.


lvonk (changes requested):
I suddenly realise this won't work if the Aggregate is snapshotted.

PR #308, opened 2022-02-11
[feature] define after_initialize callback in sequent command

crypto-zhiming:
Define callback `after_initialize` in `Sequent::Command`. We can use this callback to do more things, such as set default value for command attributes.

```ruby
class AfterInitCommand < Sequent::Command
  attrs flag: Boolean

  after_initialize do
    @flag = true if @flag.nil?
  end
end
```

### DIFF
--- a/lib/sequent/core/command.rb
+++ b/lib/sequent/core/command.rb
@@ -31,9 +31,13 @@ module Sequent
 
       attrs created_at: DateTime
 
+      define_model_callbacks :initialize, only: :after
+
       def initialize(args = {})
         update_all_attributes args
         @created_at = DateTime.now
+
+        _run_initialize_callbacks
       end
 
       def self.inherited(subclass)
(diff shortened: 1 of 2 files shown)


bforma:
I'm not sure if this is the way to go for 'default' values.

We could also introduce the following:

```ruby
MyCommand < Sequent::Command
  attr flag: Boolean, default: true
end
```

lvonk:
> I'm not sure if this is the way to go for 'default' values.
> 
> We could also introduce the following:
> 
> ```ruby
> MyCommand < Sequent::Command
>   attr flag: Boolean, default: true
> end
> ```
> 

I think `after_initialize` is just another active model callback we can support (like we already do with validations). Using `after_initialize` also has access to other attributes in the object. 

But we could also implement the `default` suggestion, that is probably more elegant to set default values.

lvonk (approved):

PR #311, opened 2022-02-14
feature/load aggregate for snapshotting

emielhagen:
- Introduce time specific loading of aggregates, ignoring snapshots
- Spec for testing time specific aggregate streaming

### DIFF
--- a/lib/sequent/core/aggregate_repository.rb
+++ b/lib/sequent/core/aggregate_repository.rb
@@ -58,9 +58,8 @@ module Sequent
       end
 
       # Optimizing for loading lots of events and ignore snapshot events
-      def load_aggregate_for_snapshotting(aggregate_id, _clazz = nil, load_until: nil)
-        fail ArgumentError, 'aggregate_id is required' unless aggregate_id
-        return [] if aggregate_id.blank?
+      def load_aggregate_for_snapshotting(aggregate_id, clazz = nil, load_until: nil)
+        fail ArgumentError, 'aggregate_id is required' if aggregate_id.blank?
 
         stream = Sequent
           .configuration
@@ -72,8 +71,12 @@ module Sequent
           .configuration
           .event_store
           .stream_events_for_aggregate(aggregate_id, load_until: load_until) do |event_stream|
-            aggregate.stream_from_history(event_stream)
-          end
+          aggregate.stream_from_history(event_stream)
+        end
+
+        if clazz
+          fail TypeError, "#{aggregate.class} is not a #{clazz}" unless aggregate.class <= clazz
+        end
         aggregate
       end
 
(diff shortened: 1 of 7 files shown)


lib/sequent/core/aggregate_root.rb
```diff
@@ -80,6 +71,35 @@ def load_from_history(stream, events)
...
+        events.each { |event| apply_event(event) }
+      end
+
+      def self.stream_from_history(stream)
+        aggregate_root = allocate
+        aggregate_root.initialize_for_streaming(stream)
+        aggregate_root
+      end
+
+      def initialize(id)
```
lvonk, line 94: Can you move the `initialize` back to its original position? I think it is good to have that as first instance method and reduces the number of changes.

lib/sequent/core/event_store.rb
```diff
@@ -51,7 +51,34 @@ def commit_events(command, streams_with_events)
...
+        fail ArgumentError, 'no stream found for this aggregate' if stream.blank?
+
+        q = Sequent
+          .configuration
+          .event_record_class
+          .where(aggregate_id: aggregate_id)
+          .where.not(event_type: Sequent.configuration.snapshot_event_class.name)
+          .order(:sequence_number)
+        q = q.where('created_at < ?', load_until) if load_until.present?
+        fail ArgumentError, 'no events for this aggregate' if q.blank?
```
lvonk, line 67: If there are lots of events this will do a possible expensive count query first. Perhaps we can track this in the `in_batches` loop to see if there were any events at all.

lvonk (changes requested):
Nice addition. I think it is good to add the type check.

lib/sequent/core/event_store.rb
```diff
@@ -51,7 +51,36 @@ def commit_events(command, streams_with_events)
       end
 
       ##
-      # Returns all events for the aggregate ordered by sequence_number
+      # Returns all events for the aggregate ordered by sequence_number, disregarding snapshot events
+      #
+      def stream_events_for_aggregate(aggregate_id, load_until: nil)
```
lvonk, line 56: I think it is good to add the `&block` parameter as well since that is required.

lib/sequent/core/event_store.rb
```diff
@@ -51,7 +51,36 @@ def commit_events(command, streams_with_events)
       end
 
       ##
-      # Returns all events for the aggregate ordered by sequence_number
+      # Returns all events for the aggregate ordered by sequence_number, disregarding snapshot events
```
lvonk, line 54: I think it is also good to explain the use case for this method here.

lib/sequent/core/aggregate_repository.rb
```diff
@@ -57,6 +57,29 @@ def load_aggregate(aggregate_id, clazz = nil)
         load_aggregates([aggregate_id], clazz)[0]
       end
 
+      # Optimizing for loading lots of events and ignore snapshot events
```
lvonk, line 60: I think it is also good to explain the use case for this method here.

lvonk (commented):
Just two more suggestions for improvements 👍🏼

lvonk (approved):
Looks good, great work 👍🏼

PR #315, opened 2022-04-08
command_record.command producing wrong constant name exception

gen1321:
There is a bug with `Sequent::Core::CommandRecord.last.command`.
It will fail with `wrong constant name` if you have a class with a namespace, for example: `Foo::Bar`

I believe it happens because Class.const_get should accept string instead of symbol


lvonk (approved):
Good catch 👍🏼 . It would be really appreciated if you can create a spec for this as well 🙏🏼

gen1321:
@lvonk done :)

lvonk (approved):

lvonk:
@gen1321 Thanks!

PR #316, opened 2022-04-11
Auto-register message classes for included modules & super classes

bforma:


lib/sequent/core/aggregate_root.rb
```diff
@@ -40,6 +40,8 @@ class AggregateRoot
       include Helpers::AutosetAttributes
       include SnapshotConfiguration
 
+      message_base_class Sequent::Core::Event
```
lvonk, line 43: I would prefer to not configure this here, but to keep in the MessageHandler as a Hash or something. Reason is that it is quite out of context here imo.

lvonk (commented):
I like the fact that we can match on a `is_a` kind of style.  But I am not sure about doing it statically. 

Consider this code:

```ruby
class FooEvent < Sequent::Event; end

module BarEvent; end

class FooProjector < Sequent::Projector
  on BarEvent do
  end
end

FooEvent.include BarEvent
```

Perhaps a bit farfetched but, due to the static registering of the `on` block, this projector will not be executed in this scenario right?

Another idea is to introduce an extra param with `on` like:

```ruby
on  FooEvent, matcher: :is_a do |event|
end
```

This prevents needing todo a `is_a` check for all events and does not introduce any unwanted messages being handled.

This also allows for extending into other matchers were `is_a` can be a sequent provided one.

What do you think?

bforma:
> I like the fact that we can match on a `is_a` kind of style. But I am not sure about doing it statically.
> 
> Consider this code:
> 
> ```ruby
> class FooEvent < Sequent::Event; end
> 
> module BarEvent; end
> 
> class FooProjector < Sequent::Projector
>   on BarEvent do
>   end
> end
> 
> FooEvent.include BarEvent
> ```
> 
> Perhaps a bit farfetched but, due to the static registering of the `on` block, although valid ruby this projector will not be executed in this scenario right?

Yes good point. I would expect that event class (and module) definitions are loaded prior to loading message handlers (you need the class reference anyway).

I've taken this approach because looking up the handler blocks for an `event` instance is now an `O(1)` operation (same as it was before).
 
> Another idea is to introduce an extra param with `on` like:
> 
> ```ruby
> on  FooEvent, matcher: :is_a do |event|
> end
> ```
> 
> This prevents needing todo a `is_a` check for all events and does not introduce any unwanted messages being handled.
> 
> This also allows for extending into other matchers were `is_a` can be a sequent provided one.
> 
> What do you think?

I like the idea of 'matchers' and the possibility of extension by Sequent users. However, I don't think we can make that an `O(1)` operation. It will probably become `O(n)` because you have to run through all the registered event classes/modules and stop when there is a match. It's in-memory of course, but still..

lib/sequent/core/aggregate_root.rb
bforma: Yes, we only have `EventHandler` and `CommandHandler`. I'm going to give it another try with 'runtime' checking. We probably won't need this then.

bforma:
Closing in favour of #317

PR #317, opened 2022-04-12
Sequent 5.0

bforma:
# Sequent 5.0

Sequent 5.0 introduces several advanced features for `MessageHandler`s (ie. `Sequent::AggregateRoot`, `Sequent::Projector`, `Sequent::Workflow` and `Sequent::CommandHandler`), such as:

- message matching
- processing options

## Message matching

Introduces a backward compatible way to allow more advanced matching of messages in `on` definitions. For example:

```ruby
module MyModule; end

class Money < Sequent::ValueObject
  attrs cents: Integer, currency: String
end

class MyEvent < Sequent::Event
  include MyModule

  attrs some_attribute: String,
        amount: Money
end

class MyExcludedEvent < Sequent::Event
  include MyModule
end

class MyWorkflow < Sequent::Workflow
  on MyEvent do |event|
    # you can keep matching on class name
  end

  on is_a(MyModule) do |event|
    # matches any event whose class includes MyModule
  end

  on is_a(MyModule, except: MyExcludedEvent) do |event|
    # matches any event whose class includes MyModule, but not MyExcludedEvent
  end

  on is_a(Sequent::Event) do |event|
    # matches any event whose super class is Sequent::Event
  end

  on any do |event|
    # matches any event
  end

  on any(except: MyExcludedEvent) do |event|
    # matches any event except MyExcludedEvent
  end

  on has_attrs(MyEvent, sequence_number: gt(100)) do |event|
    # matches events of class MyEvent with a sequence number greater than 100
  end

  on has_attrs(MyEvent, amount: {cents: gt(100), currency: neq('USD')}) do |event|
    # matches events of class MyEvent and have an amount of cents greater than 100 and a currency not equal to USD
  end

  on has_attrs(is_a(MyModule), some_attribute: eq('some value')) do |event|
    # matches events that include MyModule and have some_attribute with a value of 'some value'
  end

  on has_attrs(MyEvent, some_attribute: 'some value') do |event|
    # eq can also be omitted, since it's the default matcher of an attr value
  end
end
```

Some notable changes/fixes:

- `on MyEvent, MyEvent` will now raise an error stating the duplicate arguments
- `on` without arguments will now raise an error

The message matcher DSL methods (such as `is_a`, `has_attrs` and `any`) are automatically loaded when you `include Sequent::Core::Helpers::MessageHandler`.

Adding custom message matchers works as follows:

```ruby
MyMessageMatcher = Struct.new(:expected_argument) do
  def matches_message?(message)
    # return a truthy value if it matches, or falsey otherwise.
  end

  def to_s
    "my_message_matcher(#{Sequent::Core::Helpers::MessageMatchers::ArgumentSerializer.serialize_value(expected_value)})"
  end
end

Sequent::Core::Helpers::MessageMatchers.register_matcher :my_message_matcher, MyMessageMatcher
```

Which can be used as follows:

```ruby
class MyWorkfow < Sequent::Workflow
  on my_message_matcher('some constraint') do |event|
    # ...
  end
end
```

## Processing load-time options

```ruby
class MyBaseWorkflow < Sequent::Workflow
  option :deduplicate_on do |matchers, attributes|
    # This block is called for each defined `on` block that provides the corresponding option.
    #
    # In this example this is only once (for `on MyEvent, deduplicate_on: %i[aggregate_id]`, not for `on OtherEvent`).
    #
    # The first argument to this block is a list of message matchers (as defined in the `on` definition).
    # The second argument is the value specified as an argument to the option in the `on` definition.

    matchers == [Sequent::Core::Helpers::MessageMatchers::InstanceOf.new(MyEvent)] # true
    attributes == %i[aggregate_id] # true
  end
end

class MyWorkflow < MyBaseWorkflow
  on MyEvent, deduplicate_on: %i[aggregate_id] do |event|
    # ...
  end

  on OtherEvent do |event|
    # ...
  end
end
```


lib/sequent/core/helpers/message_router.rb
```diff
@@ -0,0 +1,44 @@
...
+          end
+        end
+
+        ##
+        # Returns a set of handlers that match the given message, or an empty set when none match.
+        #
+        def match_message(message)
+          @routes
+            .reduce(NO_MATCH.call) do |memo, (message_class, handlers)|
+              memo = memo.merge(handlers) if message.is_a?(message_class)
```
lvonk, line 30: This is a breaking change so we should make this optional somehow. There are several options: 

1. Put behind a configuration settings
```ruby
 memo = if Sequent.configuration.use_is_a_matcher
  memo.merge(handlers) if message.is_a?(message_class)
else
  memo.merge(handlers) if message == message_class
end
```

2. Keep the current behavior the default and implement a `matcher`  option that  you can pass  in the `on` block like we discussed in #316 

I think I prefer option 2, to implement this without an extra config item,  and for now use the `is_a` as an extra argument when registering the event using the `on` bock.

Maybe there are other options?

lvonk (commented):
Really nice and elegant with the dispatcher and router. This is however a breaking change, so we need to be backward compatible. I described some options we can discuss, but I am interested in other options you might have as well.

lib/sequent/core/helpers/message_router.rb
bforma: Some shower thoughts on alternatives:

```ruby
module Matchers
  EqualsClass = Struct.new(:expected_class, keyword_init: true) do
    def matches_message?(message)
      message.class == expected_class
    end
  end

  IsA = Struct.new(:expected_class, keyword_init: true) do
    def matches_message?(message)
      message.is_a?(expected_class)
    end
  end
end
```

```ruby
module MessageMatchers
  module ClassMethods
    def is_a?(message_class)
      Matchers::IsA.new(expected_class: message_class)
    end
  end

  def self.included(host_class)
    host_class.extend(ClassMethods)
  end
end

class MyHandler
  include MessageHandler
  include MessageMatchers

  on SpecificEvent, is_a?(EventModule) do |message|
    # ...
  end
end
```

And `SpecificEvent` will automatically be wrapped in a `Matchers::EqualsClass` for backward compatibility.

Alternative DSL:

```
class MyHandler
  include MessageHandler

  on SpecificEvent, is_a: EventModule do |message|
    # ...
  end
end
```

Where we convert the value of `is_a` to a `Matchers::IsA` (could also be an array for multiple like: `on SpecificEvent, is_a: [SomeModule, OtherModule]`).

We could also provide a built-in `HasAttrs` matcher, used like this:

```ruby
class MyHandler
  include MessageHandler

  on SpecificEvent, attrs: {aggregate_type: 'MyType'} do |message|
    # ...
  end
end
```

Either way this opens up extensibility for Sequent users, allowing them to write custom 'matchers', like so:

```ruby
class MyHandler
  include MessageHandler

  class CustomMatch
    def matches_messages?(message)
      # ...
    end
  end

  class << self
    def custom_match
      CustomMatch.new
    end
  end

  on custom_match do |message|
    # ...
  end
end
```

bforma: I've let this have some more thoughts and I would prefer the following solution:

```ruby
class MyHandler
  include MessageHandler

  on MyEvent do |event|
    # ...
  end

  on is_a(MyModule) do |event|
    # ...
  end

  on has_attrs(OtherEvent, attr: 'value', nested: {numeric: greater_than(0)}) do |event|
    # ...
  end
end
```

Advantages:

- The `on` method  signature would change from `on(*event_classes)` to `on(*message_matchers)` and we can automatically wrap a `Class` argument in a `class_equals` matcher for full backward compatibility.
- Sequent can provide built-in matchers for common cases.
- Sequent users can write their own matchers for proprietary cases.
- Matchers can be nested for more complex cases.
- No extra configuration needed.

lvonk: Nice one indeed!

lib/sequent/core/helpers/matchers/class_matcher.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Sequent
+  module Core
+    module Helpers
+      ClassMatcher = Struct.new(:expected_class, keyword_init: true) do
```
lvonk, line 6: Why not use a normal class?

```ruby
class ClassMatcher
  def initialize(expected_class:)
  end
  def matches_message?(message)
     message.is_a?(expected_class)
   end
end
```

lib/sequent/core/helpers/message_router.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+require_relative './matchers/matchers'
+
+module Sequent
+  module Core
+    module Helpers
+      class MessageRouter
+        NO_MATCH = -> { Set.new }
```
lvonk, line 9: Should this really be a Set? Might, in theory, give unexpected results?

lvonk (commented):
I think this is indeed the way to go. We also need to update the documentation for this.

lib/sequent/core/helpers/matchers/class_matcher.rb
bforma: Structs have the advantage of 'by value' comparison. They are used as keys in a hash, so existing values are overwritten by default. For example:

```
on is_a(MyModule), is_a(MyModule) do |message|
  # ...
end
```

Would result in only a single entry in the hash.

lib/sequent/core/helpers/message_router.rb
bforma: I decided to use a `Set` here in order to ensure a unique collection of handlers.

The documentation of the `MessageHandler` class already states this:

> The order of which handler block is executed first is not guaranteed.

If we want guaranteed order, we could use an `OrderedSet`.

bforma: We could also prevent users from making 'mistakes' like this by validating that provided message matchers should be unique, otherwise raise a descriptive error.

lvonk: This is not correct I think since registering two different on blocks on the same class will overwrite the first.

```ruby
on FooEvent do
  # do something
end

on FooEvent do
  # do something else
end
```

At runtime this will only execute the last one registered right? This breaks current behavior. I'd prefer to keep that the same.

bforma: Both handlers (read: blocks) will be executed in this case.

mvandiepen (approved):
Elegant DSL 👍

lvonk (approved):
Looks good!

docs/docs/concepts/advanced.md
```diff
@@ -138,3 +138,77 @@ Command: SendInvoice resulted in 2 events
...
+  def matches_message?(message)
+    # return a truthy value if it matches, or falsey otherwise.
+  end
+
+  def matcher_description
+    "my_matcher(#{expected_argument})"
+  end
+end
+
+Sequent::Core::Helpers::MessageMatchers.define_matcher :my_matcher, MyMatcher
```
lvonk, line 199: `register_matcher`

lib/sequent/core/helpers/message_matchers/has_attrs.rb
```diff
@@ -32,12 +32,12 @@ def matches_attrs?(message, expected_attrs, path = [])
...
-            if expected_value.is_a?(Hash)
-              matches_attrs?(message, expected_value, path)
-            else
-              expected_value = AttrMatchers::Equals.new(expected_value) unless expected_value.respond_to? :matches_attr?
-              expected_value.matches_attr?(message.attributes.dig(*path))
-            end
+            return matches_attrs?(message, expected_value, path) if expected_value.is_a?(Hash)
+
+            expected_value = AttrMatchers::Equals.new(expected_value) unless expected_value.respond_to?(:matches_attr?)
+            value = path.reduce(message) { |memo, p| memo.public_send(p) }
```
bforma, line 38: Nice one! 👍

Issue #320, opened 2022-04-26
Getting "TypeError: can't quote Array" when loading multiple aggregates


victorfeijo:
Hey everyone! 👋🏽 

I've used Sequent for quite a while at my previous job and had a really great experience! Recently I started a new project and we decided to use Sequent for our wallet architecture. While creating a new aggregate for `Credits` today, I've realized that when I try to load multiple aggregates in a single repository call I'm getting the following error:

```ruby
> wallet = Sequent.aggregate_repository.load_aggregate(["15a28365-2ebd-4727-9b85-9c0a600c0d09"], Wallets::Wallet)
TypeError: can't quote Array
from /Users/victorfeijo/.rvm/gems/ruby-3.1.1/gems/activerecord-7.0.2/lib/active_record/connection_adapters/abstract/quoting.rb:25:in `quote'
```

If I remove the array and try to load the aggregate directly I can do it without problems but seems like when trying to load multiple aggregates it's raising a `TypeError`. According to the docs described [here](https://www.sequent.io/docs/concepts/aggregate-repository.html#loading-aggregateroots), we should be able to load multiple aggregates. Is this a bug or something that I'm missing?

Our current project build:

Ruby: 3.1.1
Rails: 7.0.0
ActiveRecord: 7.0.2

Let me know if that is helpful. Thank you!

victorfeijo:
Solved by using `load_aggregates` instead of `load_aggregate`. Just created a PR to update the docs as well.

PR #321, opened 2022-04-26
Update aggregate-repository.md docs

victorfeijo:
Update docs to use `load_aggregates` instead of `load_aggregate` for loading multiple aggregates in a single call. More info at the following issue: https://github.com/zilverline/sequent/issues/320


lvonk (approved):

lvonk:
Thanks 🙏🏼

PR #322, opened 2022-05-09
Introducing: has_attrs matcher

bforma:


lib/sequent/core/helpers/message_matchers/has_attrs.rb
```diff
@@ -25,13 +25,21 @@ def matcher_description
...
+          def matches_attrs?(message, expected_attrs, path = [])
+            expected_attrs.all? do |(name, expected_value)|
+              matches_attr?(message, expected_value, path.dup.push(name))
+            end
+          end
+
+          def matches_attr?(message, expected_value, path)
+            if expected_value.is_a?(Hash)
+              matches_attrs?(message, expected_value, path)
+            else
```
lvonk, line 37: `if.. elsif... else` prevents an extra nesting. A bit more readable imo

lib/sequent/core/helpers/argument_serializer.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Sequent
+  module Core
+    module Helpers
+      class ArgumentSerializer
+        class << self
+          def serialize_value(value)
```
lvonk, line 8: This class confused me a bit since it has a generic name but is specific to Matchers. Perhaps move this to a Matcher helper or anything?

lib/sequent/core/helpers/message_matchers/any.rb
```diff
@@ -4,13 +4,21 @@ module Sequent
...
     module Helpers
       module MessageMatchers
-        class Any
-          def matches_message?(_message)
-            true
+        Any = Struct.new(:opts) do
+          include ExceptOpt
+
+          def matches_message?(message)
+            true unless excluded?(message)
```
lvonk, line 11: Isn't this the same as: `!excluded(message)`

lib/sequent/core/helpers/message_matchers/except_opt.rb
```diff
@@ -0,0 +1,25 @@
...
+
+module Sequent
+  module Core
+    module Helpers
+      module MessageMatchers
+        module ExceptOpt
+          private
+
+          def excluded?(message)
+            return false unless except
```
lvonk, line 11: If we add `compact` below we can remove this early return.

lvonk (approved):
Few comments, ott looks good.

lvonk (approved):

PR #323, opened 2022-05-12
Introducing: message handler options

bforma:
Example usage:

```ruby
class MyEventHandler
  include Sequent::Core::Helpers::MessageHandler

  option :my_option do |matcher, argument|
    # This block is called only during initialisation, for each matcher declared in an `on` definition.

    argument == :my_value # true
  end

  on MyEvent, my_option: :my_value do |event|
    # ...
  end
end
```

Can be useful for cases like this:

```ruby
module DeduplicableWorkflow
  extend ActiveSupport::Concern

  included do
    option :deduplicate_on do |matcher, attributes|
      # Stick matcher + attributes in some registry somewhere, to be dealt with later.
    end
  end
end

class BaseWorkflow < Sequent::Workflow
  include DeduplicableWorkflow
end

class MyWorkflow < BaseWorkflow
  on MyEvent, deduplicate_on: %i[aggregate_id] do |event|
    # ...
  end  
end
```

Registered options are scoped per class hierarchy, some in the above example, all workflows extending from `BaseWorkflow` support the `deduplicate_on` option in `on` definitions. Classes in other hierarchies (like a `Sequent::Projector`) don't have this option (but can register their own options of course).

### DIFF
--- a/docs/docs/concepts/advanced.md
+++ b/docs/docs/concepts/advanced.md
@@ -210,7 +210,7 @@ For a list of supported built-in message matchers, see: https://www.rubydoc.info
 
 For a list of supported built-in attr matchers, see: https://www.rubydoc.info/gems/sequent/Sequent/Core/Helpers/AttrMatchers.
 
-## Custom message matcher
+### Custom message matcher
 
 You can also provide your own custom message matchers as follows:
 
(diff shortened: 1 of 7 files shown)


lvonk (approved):
Nice one, this does need some documentation in the (for now) advanced section on how you can use this.

PR #327, opened 2022-06-13
Pass all message matchers 

bforma:


lvonk (commented):
This looks good, but can you elaborate a bit on the reasoning behind this change?

bforma:
Yes of course. Allow me to explain.

The current approach is to call the block provided to `option` for each message matcher provided to `on`, like so:

```ruby
module DeduplicableWorkflow
  extend ActiveSupport::Concern

  included do
    option :deduplicate do |matcher, opts| # called multiple times, for each matcher defined in on blocks
      DeduplicationRegistry.register_entry(self, matcher, **opts)
    end
  end
end

class MyWorkflow < Sequent::Workflow
  on MyEvent, OtherEvent, deduplicate: {on: %i[aggregate_id]} do |event|
    # ...
  end
end
```

This will call `DeduplicationRegistry.register_entry(self, matcher, **opts)` twice (once for each message matcher provided to `on`, so for `MyEvent` and for `OtherEvent` (wrapped in a matcher)).

This way we can't do any grouping on a list of message matchers (because they are provided once at a time in the block provided to `option`).

The following does allow future implementations to do that (and we could actually use this in Jortt for improved deduplication):

```ruby
module DeduplicableWorkflow
  extend ActiveSupport::Concern

  included do
    option :deduplicate do |matchers, opts| # called only once per defined on block
      DeduplicationRegistry.register_entry(self, matchers, **opts)
    end
  end
end

class MyWorkflow < Sequent::Workflow
  on MyEvent, OtherEvent, deduplicate: {on: %i[aggregate_id]} do |event|
    # ...
  end
end
```

I would have to elaborate on the implementation of `DeduplicationRegistry.register_entry` here to explain implementation improvements, but I think it should be suffice to say that calling the block passed to `option` only once (with each message matcher passed in a list) allows developers/users of this more control over how to handle it.

PR #328, opened 2022-06-14
Store reference to EventRecord#id in Event

bforma:
When publishing.


lvonk:
Can't we use aggregate_id and sequence_number? That is what identifies an Event.

bforma:
Yes, I agree that is much better. Closing this.

PR #329, opened 2022-08-04
Add support for multiple database applications

sam-cw:
In ActiveRecord 6+ multiple databases can be used under specific roles. For example a read-replica can be used for read only roles. This requires a different database configuration structure and connection handling.

This PR allows Sequent to connect to databases configured in this way, though due to the nature of event sourcing it can only connect to a single database, which should always be the primary database.


lib/sequent/configuration.rb
```diff
@@ -107,6 +109,13 @@ def initialize
 
       self.logger = Logger.new(STDOUT).tap { |l| l.level = Logger::INFO }
       self.error_locale_resolver = DEFAULT_ERROR_LOCALE_RESOLVER
+
+      self.enable_multiple_database_support = false
+      self.primary_database_key = :primary
+    end
+
+    def can_use_multiple_databases
```
lvonk, line 117: It would be nice to rename to `can_use_multiple_databases?`

lib/sequent/support/database.rb
```diff
@@ -66,21 +101,27 @@ def self.with_schema_search_path(search_path, db_config, env = ENV['RACK_ENV'])
...
 
         if ActiveRecord::VERSION::MAJOR < 6
           ActiveRecord::Base.configurations[env.to_s] =
             ActiveSupport::HashWithIndifferentAccess.new(db_config).stringify_keys
         end
 
-        db_config[:schema_search_path] = search_path
-
-        ActiveRecord::Base.establish_connection db_config
+        if Sequent.configuration.can_use_multiple_databases
```
lvonk, line 110: Afterwards we should set the search_path back to what it was before like the previous version of this method. So in short:

```ruby
def self.can_use_multiple_databases(search_path)
  original_search_paths = get_current_search_path
  set_search_path(search_path)
  yield
ensure
  set_search_path(original_search_paths)
end
```

lib/sequent/support/database.rb
```diff
@@ -30,17 +31,51 @@ def self.read_config(env)
       end
 
       def self.create!(db_config)
-        ActiveRecord::Base.establish_connection(db_config.merge(database: 'postgres'))
-        ActiveRecord::Base.connection.create_database(db_config[:database])
+        if Sequent.configuration.can_use_multiple_databases
+          ActiveRecord.legacy_connection_handling = false
```
lvonk, line 35: I see these couple of lines on multiple places, can we maybe  extract the into a single object?

lvonk (commented):
Nice addition! Thanks a lot. Can we try to put the connection code into a single object?

lib/sequent/support/database.rb
```diff
@@ -30,17 +31,51 @@ def self.read_config(env)
...
+        else
+          ActiveRecord::Base.establish_connection(db_config.merge(database: 'postgres'))
+          ActiveRecord::Base.connection.create_database(db_config[:database])
+        end
       end
 
       def self.drop!(db_config)
-        ActiveRecord::Base.establish_connection(db_config.merge(database: 'postgres'))
-        ActiveRecord::Base.connection.drop_database(db_config[:database])
+        if Sequent.configuration.can_use_multiple_databases?
```
bforma, line 53: This block of code looks similar to that in `create`. Can be dry it up into a separate method/class?

lib/sequent/support/database.rb
```diff
@@ -66,21 +101,27 @@ def self.with_schema_search_path(search_path, db_config, env = ENV['RACK_ENV'])
...
         if ActiveRecord::VERSION::MAJOR < 6
           ActiveRecord::Base.configurations[env.to_s] =
             ActiveSupport::HashWithIndifferentAccess.new(db_config).stringify_keys
         end
 
-        db_config[:schema_search_path] = search_path
-
-        ActiveRecord::Base.establish_connection db_config
-
+        if Sequent.configuration.can_use_multiple_databases?
```
bforma, line 110: Idem.

bforma (approved):
Nice work!

Issue #219, opened 2020-01-08
command_helpers and workflow_helpers conflict


yanovitchsky:
Hello in the sequent spec helper when adding the workflow helpers it seems like the command helpers do not work anymore.
the method then_events always return an empty array
```ruby
RSpec.configure do |config|
  config.include Sequent::Test::WorkflowHelpers
  config.include Sequent::Test::CommandHandlerHelpers
end
```
As a workaround, I added a rspec configure block in the workflow spec

lvonk:
Thanks for reporting. Will look into this coming days.

lvonk:
The `WorkflowHelpers` need to use the `FakeCommandService` and register it in the Sequent config. So we need to somehow detect which spec `WorkflowHelpers` and which needs the `CommandHandlerHelpers`.

So we can:
1) document that the `WorkflowHelpers` can only be used in single specs.
2) we can include the `WorkflowHelpers` by using a tag:

```ruby
describe SendMailWorkflow, :use_workflow_helpers do
  it 'will only include Sequent::Test::WorkflowHelpers if the tag use_workflow_helpers is used'
end
```

My preference has 2. What's your preference?

gen1321:
There is another problem that fake command service from workflows helpers are leaking to other specs prevent commands from being executed, and as a result makes tests fail. It's kinda hard to spot because it not always reproducible(it happens only if your workflow specs were run before your command handler ones). To temporarily solve this issue, I added 
```
  before :each do
    .......
    Sequent.configuration.command_service = Sequent::Core::CommandService.new
    .......
  end
```

@lvonk I probably can make a PR on this(not sure about solution yet tho), would you be interested?

lvonk:
Yes that would be nice. Although I think we should solve the conflicting `WorkflowHelpers` and `CommandHandlerHelpers ` at the same time. We could have something like this (like in `EventHandlerHelpers`:

```ruby
around :each, :some_tag do |spec|
  old_config = Sequent.configuration
  Sequent::Configuration.reset
  # set fake command store
  spec.run
ensure
  Sequent::Configuration.restore(old_config)
end
```

PR #340, opened 2022-10-14
Fix conflict when using Sequent::Test::WorkflowHelpers

lvonk:
Using WorkflowHelpers could potentially and most likely break other specs
running after any spec including WorkflowHelpers. This was due to the fact
this spec altered the Sequent.configuration without restoring its original value.

To ensure this WorkflowHelpers plays nicely with other specs you
must, as of now, include it using `workflows: true` metadata:

```ruby
  RSpec.configure do |config|
    config.include Sequent::Test::WorkflowHelpers, workflows: true
  end
```


lvonk:
Fixes https://github.com/zilverline/sequent/issues/219

PR #342, opened 2022-10-19
Improve Rails / Sequent documentation

lvonk:


Meryllin (approved):
The changes look good 🚀 should a caveat be made for Rails engines as well (i.e. explicitly state the same principles should apply for Rails engines)? 🤔

lvonk:
Yes i can add a word on Rails engines

PR #343, opened 2022-10-20
Improve support for creating sequent_schema and view_schema for testing

lvonk:
Added DatabaseHelpers.maintain_test_database_schema to ease setup
sequent view_schema and sequent_schema for testing.
Prior the rake task need to be manually run. This can now be replaced
by adding

  Sequent::Test::DatabaseHelpers.maintain_test_database_schema(env: 'test')

to for instance your spec_helper.rb.

For people using database cleaner with delete or truncation strategy. You must
exclude the Sequent::Migrations::ViewSchema::Versions.table_name from being
emptied, since Sequent relies on the contents for this table to determine if tables
need to be created.

So add the following when using DatabaseCleaner:

  DatabaseCleaner.clean_with(:truncation, {except: Sequent::Migrations::ViewSchema::Versions.table_name


celinelemenn (approved):
Nice !

PR #344, opened 2022-10-26
improve maintain test schema

lvonk:
- Allow maintain_test_database_schema usage for development as well
- Fix replay_all! and allow to pass group_exponent as param

### DIFF
--- a/lib/sequent/test/database_helpers.rb
+++ b/lib/sequent/test/database_helpers.rb
@@ -3,13 +3,13 @@
 module Sequent
   module Test
     module DatabaseHelpers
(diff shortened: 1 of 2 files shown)


lib/sequent/test/database_helpers.rb
```diff
@@ -3,11 +3,13 @@
 module Sequent
   module Test
     module DatabaseHelpers
+      ALLOWED_ENS = %w[development test spec].freeze
```
mvandiepen, line 6: `ALLOWED_ENVS` :)

mvandiepen (approved):

PR #354, opened 2022-11-07
Replay events using BETWEEN predicate on aggregate_id

erikrozendaal:
Instead of using a prefix with a separate index. This will allow us to drop the `replay_events` index on `event_type,
substring(aggregate_id::text, 1, 3)` since it uses the existing `unique_event_per_aggregate` index instead.

This implementation is slower though when only a few event types need to be replayed, instead of all events. Not sure if this is a big problem or not.

### DIFF
--- a/lib/sequent/migrations/view_schema.rb
+++ b/lib/sequent/migrations/view_schema.rb
@@ -341,10 +341,11 @@ module Sequent
 
       def aggregate_id_ranges(exponent)
         fail "Exponent must be in range #{ALLOWED_REPLAY_EXPONENT}" unless ALLOWED_REPLAY_EXPONENT.include? exponent
-        all_prefixes = (0...16**exponent).map do |i|
-          prefix = "%0*x" % [exponent, i]
-          lowerbound = prefix + '00000000-0000-0000-0000-000000000000'[exponent .. -1]
-          upperbound = prefix + 'ffffffff-ffff-ffff-ffff-ffffffffffff'[exponent .. -1]
+
+        (0...16**exponent).map do |i|
+          prefix = format('%0*x', exponent, i)
+          lowerbound = prefix + '00000000-0000-0000-0000-000000000000'[exponent..-1]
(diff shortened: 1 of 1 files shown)


lvonk:
> This will allow us to drop the replay_events index on event_type, substring(aggregate_id::text, 1, 3) since it uses the existing unique_event_per_aggregate index instead.

IIRC we added the `replay_events` index because of the speed improvement it had over a `between` of uuid's. Of course this was a few years back and with an older postgres version so we should definitely recheck our old assumptions.

> This implementation is slower though when only a few event types need to be replayed, instead of all events. Not sure if this is a big problem or not.

This is the most typical scenario where you only rebuild a few projectors and therefore a limited set of event types. In my opinion we should optimize for this use case. The problem depends on how much slower it is 😅 ?

erikrozendaal:
> IIRC we added the `replay_events` index because of the speed improvement it had over a `between` of uuid's. Of course this was a few years back and with an older postgres version so we should definitely recheck our old assumptions.

Adding an `INCLUDE (event_type)` on the `unique_event_per_aggregate` table will probably return a lot of the performance, since it will not have to scan every row of the `event_records` table when only a few event types are of interest. But this does increase the size of this index by about 20 GB (currently 47 GB). The size of the `replay_events` index is 48 GB, which can be dropped.

> > This implementation is slower though when only a few event types need to be replayed, instead of all events. Not sure if this is a big problem or not.
> 
> This is the most typical scenario where you only rebuild a few projectors and therefore a limited set of event types. In my opinion we should optimize for this use case. The problem depends on how much slower it is 😅 ?

A quick "explain analyze" shows that it takes about 1 minute to run for 1/4096th of the table on the current production DB. Most of that will be waiting for disk. When data is in memory it takes less than a second.

The current implementation is a lot faster (2-10 seconds with cold cache, depending on the selectivity of the event types). But there is a continuous drag on performance to maintain the `replay_events` index.

Note that when we add data partitioning a different strategy will have to be used, based on the partitions.

PR #355, opened 2022-11-18
Use PostgreSQL stored procedures for loading and storing events

erikrozendaal:
This is in preparation of optimizing the schema to reduce space requirements and allow for table partitioning. By using the stored procedures as an interface to the database the schema can be changed and (incrementally) migrated without requiring downtime.

### DIFF
--- a/lib/sequent/support/view_projection.rb
+++ b/lib/sequent/support/view_projection.rb
@@ -1,0 +1,58 @@
+# frozen_string_literal: true
+
+require 'postgresql_cursor'
+
+module Sequent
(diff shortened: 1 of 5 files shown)


lib/sequent/support/view_projection.rb
```diff
@@ -0,0 +1,58 @@
+# frozen_string_literal: true
+
+require 'postgresql_cursor'
+
+module Sequent
+  module Support
+    class ViewProjection
```
lvonk, line 7: What is this for?

PR #358, opened 2022-12-20
Handle event within fake transaction

jankeesvw:
This ensures the after_commit is called in tests.

### DIFF
--- a/lib/sequent/test/workflow_helpers.rb
+++ b/lib/sequent/test/workflow_helpers.rb
@@ -78,7 +78,9 @@ module Sequent
       end
 
       def when_event(event)
-        workflow.handle_message event
+        fake_transaction_provider.transactional do
+          workflow.handle_message event
+        end
       end
 
       def then_commands(*commands)


lvonk (commented):
Thanks, this is useful to have but I prefer this being optional perhaps by passing in an optional variable to `when_event(MyEvent, run_after_commit: true)` ?

jankeesvw:
We could create a setting for this changed behavior, and maybe change the default in a major Sequent version update. 

But I think this is the correct default. It also only affects tests, whats the reason you don't want to change this behavior?

lvonk:
> We could create a setting for this changed behavior, and maybe change the default in a major Sequent version update.

This is also a good one. Reason is breaking changes, that was my main concern.

jankeesvw:
> This is also a good one (better than my suggestion :-)). Reason is breaking changes, that was my main concern.

Can you try it on your code bases? I don't think it will actually break something. 🚢 :shipit:

lvonk:
Ok, will merge. Can always introduce the flag at a later stage if needed

lvonk (approved):

PR #359, opened 2023-01-10
Add option to not cache event types

bforma:
When using Rails auto-loading, classes should not be cached.

> Bottom line: do not cache reloadable classes or modules.

Source: https://guides.rubyonrails.org/autoloading_and_reloading_constants.html#reloading-and-stale-objects

The `Sequent::Core::EventStore` caches event classes by default (which is good, because it removes the need for unnecessary slow constant lookups). When using Rails in development we want to disable this, otherwise the `on` blocks registered in objects such as `Sequent::AggregateRoot`s aren't called properly (because the cached (old) event class does not match the one registered in a message handler).

Usage:

```ruby
Sequent.configure do |config|
  config.event_store = Sequent::Core::EventStore.new(cache_event_types: !Rails.env.development)
end
```

The `cache_event_types` argument has a default value of `true` for backward compatiblity.

### DIFF
--- a/lib/sequent/core/event_store.rb
+++ b/lib/sequent/core/event_store.rb
@@ -26,8 +26,21 @@ module Sequent
         end
       end
 
-      def initialize
-        @event_types = ThreadSafe::Cache.new
+      ##
+      # Disables event type caching (ie. for in development).
+      #
+      class NoEventTypesCache
+        def fetch_or_store(event_type)
+          yield(event_type)
+        end
+      end
+
+      def initialize(cache_event_types: true)
+        @event_types = if cache_event_types
+                         ThreadSafe::Cache.new
+                       else
+                         NoEventTypesCache.new
+                       end
       end
 
       ##
--- a/spec/lib/sequent/core/event_store_spec.rb
+++ b/spec/lib/sequent/core/event_store_spec.rb
@@ -195,7 +195,7 @@ describe Sequent::Core::EventStore do
         )
         stream, events = event_store.load_events(aggregate_id)
         expect(stream).to be
-        expect(events).to eq([TestEventForCaching.new(aggregate_id: aggregate_id, sequence_number: 1)])
+        expect(events.first).to be_kind_of(TestEventForCaching)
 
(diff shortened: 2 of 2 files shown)


lvonk:
This is literally something we ran into recently and realized that reloading was the culprit 🙏🏼 . We are now investigating to not reload anything in `app/domain`, but that is far from ideal. 
I think this also applies to Commands by the way which are cached in `Sequent::Core::Commands`. So maybe we need a more global setting for this?

bforma:
> This is literally something we ran into recently and realized today that reloading was the culprit 🙏🏼 for this sometimes weird behavior . We are now investigating to not reload anything in app/domain, but that is far from ideal.

Cool. We currently reload everything (also `app/domain`) and it works fine (now).

> I think this also applies to Commands by the way which are cached in Sequent::Core::Commands. So maybe we need a more global setting for this?

That feels also broken yes, but it hasn't hurt me yet. I've looked into it briefly and I think we can replace the `self.inherited` implementations with `ActiveSupport::DescendantsTracker` (which I think is auto-loading compatible). I'll come back to this.

lvonk:
I will test this PR on our code base tomorrow and see if it also resolves our issue

lvonk:
One of the previous issues was with `ActiveSupport::DescendantsTracker` (or it's predecessor) was that it traversed the entire ObjectSpace making it very slow. 
But I believe that has been fixed for quite a while. We should indeed replace `Commands` with `DescendantsTracker`

bforma:
> I will test this PR on our code base tomorrow and see if it also resolves our issue

Any luck?

lvonk:
Yeah, no luck. When using this branch I get a weird error it can't find the `database.yml` 🤔 . Can you rebase with current master and I will try again.

bforma:
@lvonk I've updated with `master` and fix any failing specs, can you try again?

lvonk:
Yes I did that, unsure if this is due to background tasks using shoryuken or not. But currently too busy to check if that is the cause...

PR #362, opened 2023-01-16
Use `ActiveSupport::DescendantsTracker`

bforma:
Use `ActiveSupport::DescendantsTracker` for fast subclass lookups.

> This module provides an internal implementation to track descendants which is faster than iterating through ObjectSpace.

Also see https://api.rubyonrails.org/classes/ActiveSupport/DescendantsTracker.html.

The following classes have been kept and use `.descendants` for fast lookups internally:

- `Sequent::Core::Commands`
- `Sequent::Core::AggregateRoots`
- `Sequent::Core::Projectors`
- `Sequent::Core::Workflows`

### DIFF
--- a/lib/sequent/core/aggregate_root.rb
+++ b/lib/sequent/core/aggregate_root.rb
@@ -39,13 +39,9 @@ module Sequent
       include Helpers::MessageHandler
       include Helpers::AutosetAttributes
       include SnapshotConfiguration
+      extend ActiveSupport::DescendantsTracker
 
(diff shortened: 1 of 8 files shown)


lvonk (commented):
I'd prefer to keep these wrapper classes, not only because we use them but also because they provide searching by name. This is handy when locating a command by name. E.g `Sequent::Core::Commands.find(payload['command_name'])`.

bforma:
Ah, let's not deprecate them then.

bforma:
Can this be merged?

PR #370, opened 2023-03-02
Fix: Change db:migrate error message ENV

coalest:
I continued to receive this error message about wrapping `db:migrate` even after creating a Rake task identical to the one in the error message. Upon investigation, looks like this `ENV` variable is supposed to match the one a few lines above.


lvonk (approved):
Good catch! Thanks!

PR #371, opened 2023-04-07
Feat/guide improvements

anyanuttzilverline:
Some changes to ease the process of following the guide:
- Added fallback setting of SEQUENT_ENV to 'development' in `boot.rb`
- Set YAML loader to load aliases in `app/database.rb`
- update `on PostAuthorChanged` in `post.rb` to use field author_agregate_id rather than author
- fix minor spelling error in directory mention: `app/view/layout.erb` -> `app/views/layout.erb`


lvonk (approved):
Thanks! 🙏🏼

PR #373, opened 2023-04-12
Feat/docs improvements

anyanuttzilverline:


lvonk (commented):
Thanks, this is great, bye bye denglish.... 😆 . Only one thing on the rename of an event in an example.

PR #381, opened 2023-07-03
Reset the column information on all tables in the plan when migrated online

lvonk:
If rake sequent:migrate:offline is called in the same memory space like:
`rake sequent:migrate:online sequent:migrate:offline` it fails.
This scenario is typically done in development environments in which there
is no need for downtime.


sam-cw (approved):
Nice

PR #388, opened 2023-09-01
Translate custom errors by error_locale_resolver

mvandiepen:
`command.valid?` fills the error messages. So wrapping that with `I18n.with_locale` ensures those messages are translated according to the `error_locale_resolver`.

Previously tried to ensure this through a command_middleware, like: 

```ruby
class EnsureCommandErrorLocale
  def self.call(command, &block)
    I18n.with_locale(Sequent.configuration.error_locale_resolver&.call || I18n.locale) do
      command.valid?
    end

    block.call
  end
end
```

But the `command.valid?` called in `Sequent::Core::CommandService#process_command` will override the messages again, reverting it back to the default `I18n.locale`. Hence this PR.

### DIFF
--- a/lib/sequent/core/command_service.rb
+++ b/lib/sequent/core/command_service.rb
@@ -71,7 +71,9 @@ module Sequent
 
         filters.each { |filter| filter.execute(command) }
 
-        fail CommandNotValid, command unless command.valid?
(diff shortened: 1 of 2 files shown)


lvonk (commented):
lgtm, can you please squash the commits?

PR #393, opened 2023-10-22
Introduce autoregistering command and event handlers

lvonk:
Setting up (large) Sequent configs can become quite tedious since all CommandHandlers and EventHandlers need to be configured explicitly.

By setting `enable_autoregistration` to `true` this will no longer
be needed and all `Sequent::CommandHandler`s, `Sequent::Projector`s and
`Sequent::Workflow`s not marked as *abstract_class* or *skip_autoregister*
are *appended* to already configured `command_handlers` and
`event_handlers`.

To use:

Add the following code to the sequent initializer:

```ruby
Sequent.configure do |config|
  config.enable_autoregistration = true
end
```

**Note for Rails users:**
In order for this to work Sequent will eager load all code **before** Sequent initializer runs since otherwise the `command_handlers` and `event_handlers` are still empty after initialization since normally eager_loading runs **after** all initializers ran.

This might cause other issues if initializers after sequents initializer depend on the fact the code is **not** eager loaded yet. If that is the case ensure that the Sequent initializer runs last.

This change will require a major version bump as we have introduced (some small) breaking changes.

TODO:
- [x] Test in Rails (add Rails app in integration-specs)
- [x] ~~Fix `MessageRouter` and `MessageDispatcher` for Rails with regards to https://github.com/zilverline/sequent/pull/359~~
- [x] Documentation


mvandiepen (approved):
Nice addition. 👍

docs/docs/concepts/configuration.md
```diff
@@ -1,42 +1,58 @@
...
 
+Sequent 6.0.2 introduced autoregistration of `command_handlers` and `event_handlers` via
+setting `enable_autoregistration` to `true`.
+Autoregistered classes will be appended to any already manually registered `command_handlers` and `event_handlers`. If
+Sequent detects duplicates it will currently fail.
+When setting `enable_autoregistration` to `true` in your `initializer`
+any [CommandHandlers](command-handler.html), [Projectors](projector.html) and [Workflows](workflow.html) are
+automatically registered
+in your Sequent configuration. When you have base classes that you don't want to register your self you can
+set `self.abstract_class = true` for these classes. Another option to skip autoregistration is to set
```
mvandiepen, line 54: "When you have base classes that you don't want to register your self..." should probably say: "When you have base classes that you don't want to have automatically registered..."

mvandiepen (approved):
Small nitpick in docs.

lvonk:
@bforma would you mind giving this a test run before we merge this in? All works in our apps

bforma:
I've tested this and can report the following:

- Pointed sequent to this branch.
- Had to remove the `cache_event_types` argument to `Sequent::Core::EventStore.new`.
- Removed custom initialization of event handlers and command handlers in `config/initializers/sequent.rb`.
- Configure `config.enable_autoregistration = true` in `config/initializers/sequent.rb`
- Had to configure `self.abstract_class = true` our base `Stekker::Projector`, `Stekker::CommandHandler` and `Stekker::Workflow` classes.
- Ran all tests and they are passing.
- Noticed the following in our Rails console:

```ruby
Stekker(development):001:0> Sequent.configuration.command_handlers
[
    [0] #<Sequent::Core::AggregateSnapshotter:0x000000029b45c1d8>,
    [1] #<Imbalance::PTUCommandHandler:0x000000029b45c138>
]

Stekker(development):002:0> Sequent.configuration.event_handlers
[
    [0] #<Imbalance::PTUProjector:0x000000029b45bc10 @persistor=#<Sequent::Core::Persistors::ActiveRecordPersistor:0x000000029b45bb48>>,
    [1] #<SmartCharging::ExecuteEngineWorkflow:0x000000029b45b0a8>
]
```

When running the same two commands in our Rails console on `master` I see this:

```ruby
Stekker(development):001:0> Sequent.configuration.command_handlers
[
    [0] #<Imbalance::PTUCommandHandler:0x000000016a690620>
]

Stekker(development):002:0> Sequent.configuration.event_handlers
[
    [0] #<Imbalance::PTUProjector:0x000000016a81e618 @persistor=#<Sequent::Core::Persistors::ActiveRecordPersistor:0x000000016a81e578>>,
    [1] #<SmartCharging::ExecuteEngineWorkflow:0x000000016a81ce08>
]
```

The difference seems that `Sequent::Core::AggregateSnapshotter` is now also automatically registered as a "command handler". Is that what we want/need?

lvonk:
Yes the `Sequent::Core::AggregateSnapshotter` is automatically registered if you set `enable_autoregistration = true`. Imo this makes sense otherwise you still need to manually register if you use snapshotting and have `enable_autoregistration` enabled.

PR #398, opened 2023-11-15
Add rake tasks to check for pending/running migrations

mvandiepen:


lib/sequent/rake/migration_tasks.rb
```diff
@@ -91,12 +96,28 @@ def register_tasks!
...
             desc 'Prints the current version in the database'
-            task current_version: ['sequent:init', :init] do
-              ensure_sequent_env_set!
+            task current_version: [:connect_to_db] do
+              puts "Current version in the database is: #{Sequent::Migrations::Versions.current_version}"
+            end
 
-              Sequent::Support::Database.connect!(@env)
+            desc 'Returns whether a migration is currently running'
+            task has_running_migrations: [:connect_to_db] do
```
lvonk, line 104: Perhaps `check_..._migrations`

lvonk (commented):
Nice. One comment on naming the tasks. Perhaps check as prefix is better as has sort of implies returning a true or false and this tasks just putses.

PR #399, opened 2023-11-16
Ensure view_schema exists on pending/running check

mvandiepen:


lib/sequent/rake/migration_tasks.rb
```diff
@@ -33,7 +33,7 @@ def register_tasks!
 
           task connect_to_db: ['sequent:init', :init] do
             ensure_sequent_env_set!
-            Sequent::Support::Database.connect!(@env)
+            Sequent::Migrations::ViewSchema.create_view_schema_if_not_exists(env: @env)
```
lvonk, line 36: Perhaps this should be called something like sequent migrate which is called after connect? Feels a bit weird now to have a task called connect which calls this method. Perhaps something like:

```ruby
Sequent::Support::Database.connect!(@env)
Sequent::Support::Database.migrate_sequent_schema
```

This can then of course call the `create_view_schema_if_not_exists` but it would make more sense if we look at this code later.
?

lvonk (approved):

PR #403, opened 2024-01-24
Replay optimized postgres persistor fixes

erikrozendaal:
This fixes two problems in the `ReplayOptimizedPostgresPersistor`:

1. The documentation specifies that a default `aggregate_id` index is automatically added. Unfortunately the implementation was broken.
2. The index used the `hash` value of the record keys but did not handle hash collisions at all. So when the hash function collided records could be affected with different keys, causing data corruption.

This PR fixes both issues and (tries to) clean up the code.

### DIFF
--- a/lib/sequent/core/persistors/replay_optimized_postgres_persistor.rb
+++ b/lib/sequent/core/persistors/replay_optimized_postgres_persistor.rb
@@ -56,16 +56,33 @@ module Sequent
         attr_reader :record_store
         attr_accessor :insert_with_csv_size
 
-        def self.struct_cache
-          @struct_cache ||= {}
-        end
-
-        module InitStruct
+        # We create a struct on the fly to represent an in-memory record.
+        # Since the replay happens in memory we implement the ==, eql? and hash methods
+        # to point to the same object. A record is the same if and only if they point to
+        # the same object. These methods are necessary since we use Set instead of [].
+        module InMemoryStruct
+          def ==(other)
+            equal?(other)
+          end
+          def eql?(other)
+            equal?(other)
+          end
+          def hash
+            object_id.hash
+          end
           def set_values(values)
             values.each do |k, v|
               self[k] = v
             end
             self
+          end
+        end
+
+        def self.struct_cache
+          @struct_cache ||= Hash.new do |hash, record_class|
+            struct_class = Struct.new(*record_class.column_names.map(&:to_sym))
+            struct_class.include InMemoryStruct
+            hash[record_class] = struct_class
           end
         end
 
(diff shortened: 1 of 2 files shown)


lvonk (approved):

lib/sequent/core/persistors/replay_optimized_postgres_persistor.rb
```diff
@@ -287,22 +287,18 @@ def do_with_record(record_class, where_clause)
...
+              actual_value = record[k.to_sym]
+              actual_value = actual_value.to_s if actual_value.is_a? Symbol
+              if expected_value.is_a?(Array)
+                expected_value.include?(actual_value)
+              else
+                actual_value == expected_value
               end
             end
-          end.dup
+          end).dup
```
erikrozendaal, line 301: I think this `dup` only needs to happen when the records are found in an index. So I propose moving this to `Index#find`.

Also, can it just be `freeze` and let the caller `dup` if required? Or is this too much of a trap?

lib/sequent/core/persistors/replay_optimized_postgres_persistor.rb
```diff
@@ -287,22 +287,18 @@ def do_with_record(record_class, where_clause)
...
-              where_clause.all? do |k, v|
-                expected_value = v.is_a?(Symbol) ? v.to_s : v
-                actual_value = record[k.to_sym]
-                actual_value = actual_value.to_s if actual_value.is_a? Symbol
-                if expected_value.is_a?(Array)
-                  expected_value.include?(actual_value)
-                else
-                  actual_value == expected_value
-                end
+          (@record_index.find(record_class, where_clause) || @record_store[record_class].select do |record|
```
lvonk, line 290: We still need to have `use_index?` here since `find` will always return an empty array if not found, which is `truthy`. We should add a spec for this scenario

lib/sequent/core/persistors/replay_optimized_postgres_persistor.rb
lvonk: I find this `dup` a bit weird indeed. I can't really recall why this is needed, and can't find anything in old commit messages unfortunately. It has been in there since the beginning. Since it is only the array that is dupped, not the objects in it, I am not sure what it supposed to protect against....

lvonk (changes requested):
I found a potential issue in `find_records`

lib/sequent/core/persistors/replay_optimized_postgres_persistor.rb
```diff
@@ -139,31 +139,42 @@ def update(record_class, record)
...
 
-          def find(record_class, where_clause)
-            indexes = get_indexes(record_class, where_clause)
-            return nil unless indexes.present?
-
-            normalized_where_clause = where_clause.transform_keys { |k| Persistors.normalize_symbols(k) }
-            record_sets = indexes.flat_map do |field|
-              if !normalized_where_clause.include? field
-                []
+          def find(record_class, normalized_where_clause)
```
lvonk, line 142: What does `normalized_where_clause` mean? Isn't it better to allow any valid where clause and normalize in this method?

lvonk: Or fail if it isn't normalized

erikrozendaal: This is an internal API and the normalization happens in the `find_records` method.

PR #404, opened 2024-01-26
Add migrate:dryrun task for performance testing of projectors

erikrozendaal:
This task replays all events for the selected projectors without persisting any results (so it works with a PostgreSQL read-only session).

### DIFF
--- a/lib/sequent/migrations/view_schema.rb
+++ b/lib/sequent/migrations/view_schema.rb
@@ -205,7 +205,7 @@ module Sequent
           executor.execute_online(plan)
         end
 
-        replay!(Sequent.configuration.online_replay_persistor_class.new, groups) if plan.projectors.any?
+        replay!(Sequent.configuration.online_replay_persistor_class.new, groups: groups) if plan.projectors.any?
 
         in_view_schema do
           executor.create_indexes_after_execute_online(plan)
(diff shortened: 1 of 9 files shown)


lib/sequent/migrations/view_schema.rb
```diff
@@ -283,6 +282,27 @@ def migrate_offline
         raise e
       end
 
+      def migrate_dryrun(regex, group_exponent)
+        # Complete hack to override insert_ids
+        define_singleton_method(:insert_ids) do
```
lvonk, line 287: I would move adding these methods to the rake task so they can be "accidentally" be overwritten by calling this method in the real production environment.

lib/sequent/migrations/view_schema.rb
```diff
@@ -283,6 +282,27 @@ def migrate_offline
...
+          ->(progress, done, ids) {}
+        end
+
+        persistor = Sequent.configuration.online_replay_persistor_class.new
```
lvonk, line 291: This depends on the `online_replay_persistor_class` begin used if this works or not. 
E.g. the `ActiveRecordPersistor` has no commit since that is commit when the transaction is commit.

https://github.com/zilverline/sequent/blob/29b40ffc06bdd4ca2595bad69a1256988778e3fd/lib/sequent/core/persistors/active_record_persistor.rb#L120-L122

Since this is configurable I think we should implement our own ReadonlyReplayPersistor that we set in the rake task in the sequent config to be used. Then there is no need paths this class.

Perhaps the same applies to introducting a subclass of `ViewSchema`

lvonk (changes requested):

PR #405, opened 2024-02-01
Replay optimized postgres persistor micro optimizations

erikrozendaal:
These optimizations try to reduce the number of object allocations by caching objects (like the default values of a record) or avoid allocations (by simplifying index keys or avoid constructing intermediate objects, etc).

The routing of messages based on the instance type is also optimized. Further optimization could occur by combining each projector's router into a single router, to avoid having to query each projecter if it can handle the message.

Total performance gain is about 20%-30%, depending on the projector(s) used.


lvonk (approved):
lgtm.

PR #406, opened 2024-02-02
Base event migration on minimum in-progress transaction id

erikrozendaal:
The online replay process replays all events that have a transaction id less than the minimum in-progress transaction id. The offline replay process then replays all events inserted with a greater or equal transaction id.

This avoids having to track all replayed event ids during the online process and excluding these during the offline process. This increases performance is in both processes.


lib/sequent/migrations/view_schema.rb
```diff
@@ -292,7 +291,7 @@ def ensure_valid_plan!
       def migrate_metadata_tables
         Sequent::ApplicationRecord.transaction do
           in_view_schema do
-            exec_sql([ReplayedIds.migration_sql, Versions.migration_sql].join("\n"))
+            exec_sql([Versions.migration_sql].join("\n"))
```
lvonk, line 294: Should we not drop the `ReplayedIds` in the last migration?

lvonk (commented):
Nice, It would be nice if we would also drop the ReplayedIds table.

lib/sequent/migrations/view_schema.rb
erikrozendaal: I've added a note to the changelog that this table can be dropped. We could also add a `DROP TABLE IF EXISTS ...` to Sequent but that will have to stay there "forever".

PR #419, opened 2024-09-04
Ruby 3.3

lvonk:


evsasse (approved):

erikrozendaal (approved):

lvonk:
To those interested: we found an issue in our apps connected to cloudhsm of aws after upgrading to ruby 3.3. Openssl 3 deprecated the usage of engines and by default ruby does not enable them anymore.
OpenSSL::Engine support need to be explicitly enabled by setting the compiler flag RUBY_CFLAGS=-DOSSL_USE_ENGINE (in rbenv in our case)...
So waiting to merge until this runs in production without errors.

Issue #418, opened 2024-08-20
Missing instructions to add new columns to `sequent_versions` table


evsasse:
Looks like we need to run the migration generated by this https://github.com/zilverline/sequent/blob/e55e8d526ab51e39726891b3b9aa0a7288f13193/lib/sequent/migrations/versions.rb#L11-L20, to create the new columns on the `sequent_versions` table. But we don't have official instructions for it on the CHANGELOG.md yet.

... Also, when I've generated the SQL here, I had a problem that because of the missing schema prefix ended up creating a new `sequent_versions` table, instead of updating the existing one.
- It generated `CREATE TABLE IF NOT EXISTS sequent_versions ...`
- But ideally should've been `CREATE TABLE IF NOT EXISTS view_schema.sequent_versions ...`

lvonk:
This runs (and migrates) automatically when you run the `migrate_online` task. Since it always runs with only the `view_schema` as the `search_path` it creates it in the correct schema.
Didn't it run automatically for you?

lvonk:
Hi @evsasse is this still an issue?

evsasse:
It seems to me that `migrate_online` is still problematic, that it tries to use the "status" column before it is definitely available...
https://github.com/zilverline/sequent/blob/cf7025de889f241588749bcb6be643537ac95924/lib/sequent/migrations/view_schema.rb#L188-L190

- `migrate_metadata_tables` would create the `status` column 👍.
- BUT, `ensure_valid_plan!` runs first, which calls `ViewSchema#plan`, which calls `ViewSchema#current_version`, which calls `Versions.current_version`, which executes a query that tries to use the `status` column 🤔.

lvonk:
Good catch https://github.com/zilverline/sequent/pull/422

evsasse:
> Good catch #422. Can you check if this works for you @evsasse ?

Yep, that seems to work here 👍.

PR #422, opened 2024-10-01
Ensure metada tables are migrated before checking version

lvonk:


lvonk:
Fixes https://github.com/zilverline/sequent/issues/418

evsasse (approved):
Yep, this seems to fix the problem at #418 indeed.

PR #425, opened 2024-10-08
Include default gems 'logger' and 'csv'

werdlerk:
- The [logger](https://github.com/ruby/logger) gem will no longer be part of the default gems starting from Ruby 3.5.0
- The [csv](https://github.com/ruby/csv) gem will no longer be part of the default gems starting from Ruby 3.4.0. 

Adding these gems to the gemspec should silence the warnings Ruby 3.3(.5) currently generates on every command.

```
/Users/koen/.asdf/installs/ruby/3.3.5/lib/ruby/gems/3.3.0/gems/activesupport-7.1.4/lib/active_support/logger_silence.rb:5: warning: logger was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
You can add logger to your Gemfile or gemspec to silence this warning.
/Users/koen/.asdf/installs/ruby/3.3.5/lib/ruby/gems/3.3.0/gems/sequent-7.1.1/lib/sequent.rb:5: warning: csv was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.4.0.
You can add csv to your Gemfile or gemspec to silence this warning.
```


gemfiles/README.md
```diff
@@ -0,0 +1,17 @@
+## Update Gemfiles
+The files in this directory are being used by [GitHub Actions rspec workflow](/.github/workflows/rspec.yml) to run the 
+test suite with different ActiveRecord and Ruby versions.
+
+### Update (with rbenv)
+1. Make sure you have the ruby versions listed in the [rspec workflow](/.github/workflows/rspec.yml) installed with rbenv.
+2. To update the .lock files, run:
+```shell
+RBENV_VERSION=3.0.5 BUNDLE_GEMFILE=ar_6_0.gemfile bundle install
```
lvonk, line 9: PS. This was added by an open pr in the normal README as well. https://github.com/zilverline/sequent/commit/f6b2467eb0facf2b3ee34d5362f1b11895fde6b2

werdlerk: Ah I missed that. Should I remove the `README.md` in the `gemfiles` directory? 
I see now that my commands are a bit redundant as some overwrite the already generated .lock files.

lvonk: Yeah you can remove as it is duplicate now.

PR #427, opened 2024-10-22
Sequent 8

erikrozendaal:
- Sequent now requires at least Ruby 3.2, ActiveRecord 7.1, and
  PostgreSQL 14.
- The internal database schema has been re-organized to support table
  partitioning so the events of related aggregates can be stored
  together. Partitioning is optional. You can override the
  `AggregateRoot#events_partition_key` method to return the partition
  key of an aggregate that will be used by PostgreSQL to store the
  events in the correct partition.
- Storage logic of Sequent is now done by PostgreSQL stored
  procedures. This reduces the number of round-trips necessary between
  the Ruby process and the PostgreSQL server and optimizes storage
  requirements.
- `AggregateRoot#take_snapshot!` has been replaced with
  `AggregateRoot#take_snapshot` which returns the snapshot instead of
  adding it to the uncommitted events. Snapshots can be stored using
  `EventStore#store_snapshots` method.
- The `stream_records`, `event_records`, and `command_records`
  relations are no longer tables but views on the underlying
  `aggregates`, `events`, and `commands` tables. These views take care
  of joining related tables using the correct partitioning key and
  type ids. The views are mainly for backwards compatibility with the
  `*Record` Ruby classes and can only be used for querying. For
  updates the `EventStore` (and higher-level `AggregateRepository`)
  classes should be used. New APIs have been added so there is no
  longer a need to use ActiveRecord methods directly to manage the
  event store.
- Events, commands, and snapshot are now serialized directly to a
  `JSON` or `JSONB` column. The `TEXT` column type is no longer
  supported. JSON size is minimized by extracting various repetitive
  fields (such as `aggregate_id`) into database columns.
- Snapshots are now stored into a separate `snapshot_records` table
  instead of being mixed in with events in the `events` table.  This
  also makes it possible to store snapshots without an associated
  command record. Aggregates that need snapshots are tracked in a
  separate `aggregates_that_need_snapshots` so that snapshots can be
  created more quickly by a background snapshotting process.
- The `parent`, `origin` and `children` methods of `CommandRecord` and
  `EventRecord` have been replaced by more specific methods
  (`origin_command`, `parent_event`, etc) and will always return a
  record of the appropriate type.
- Events that are updated or deleted (a rare occurrence) are archived
  into the `saved_event_records` table.
- The `id` column of `aggregates` has been removed and the primary
  key is now the `aggregate_id` column.
- The `id` column of `events` has been removed and the primary
  key are now the `aggregate_id` and `sequence_number` columns.


lvonk:
🔥 ❤️ 

As discussed:

- We should have two versions for upgrading to sequent 8: one without partitioning and one with partitioning
- A script guiding the user through the process would be really helpful
- To avoid duplication we should split `sequent_schema.sql` into tables and indexes so it can be used for both initial starters as upgraders.

lvonk:
Also, before this is merged we should release a latest 7.x.

bin/sequent
```diff
@@ -4,15 +4,25 @@
...
-def new_project(args)
-  arguments = args.dup
-  name = arguments.shift
+    Please specify a command, for example `sequent new myapp`. Available commands:
+
+      new appname                 Generate a new project from the Sequent template
+                                  project
+      generate type arguments...  Generate a new aggregate, command, or event
+
+      migrate                     Migrate a Sequent 7.1 database to Sequent 8
```
lvonk, line 19: We should also add this to the `upgrade-guide.md`.

db/sequent_8_migration.sql
```diff
@@ -0,0 +1,119 @@
+-- This script migrates a pre-sequent 8 database to the sequent 8 schema while preserving the data.
+-- It runs in a single transaction and when completed you can COMMIT or ROLLBACK the results.
+--
+-- Adjust this script to your needs (number of table partitions, etc). See comments marked with ###
```
lvonk, line 4: There is no `###` anymore in this script.

lib/sequent/core/aggregate_snapshotter.rb
```diff
@@ -24,23 +24,27 @@ class AggregateSnapshotter < BaseCommandHandler
           @last_aggregate_id,
           command.limit,
         )
-        aggregate_ids.each do |aggregate_id|
-          take_snapshot!(aggregate_id)
-        end
+        snapshots = aggregate_ids.map do |aggregate_id|
```
lvonk, line 27: `filter_map` can be used here instead of `map.compact`

lib/sequent/core/event_store.rb
```diff
@@ -174,110 +161,104 @@ def replay_events_from_cursor(get_events:, block_size: 2000,
...
+        )
       rescue ActiveRecord::RecordNotUnique
         raise OptimisticLockingError
       end
+
+      def convert_timestamp(timestamp)
+        # Since ActiveRecord uses `TIMESTAMP WITHOUT TIME ZONE`
+        # we need to manually convert database timestamps to the
+        # ActiveRecord default time zone on serialization.
+        ActiveRecord.default_timezone == :utc ? timestamp.getutc : timestamp.getlocal
```
lvonk, line 260: Perhaps we should document this that we fallback to the `:local` timezone in case it is not `utc`? Or shouldn't we use the one set in `default_timezone`?

lib/sequent/core/helpers/pgsql_helpers.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module Sequent
+  module Core
+    module Helpers
+      module PgsqlHelpers
+        def call_procedure(procedure, params)
```
lvonk, line 7: Perhaps a bit more defensive here that we expect `procedure` to have a value?

lib/sequent/core/helpers/pgsql_helpers.rb
```diff
@@ -0,0 +1,25 @@
...
+          statement = "CALL #{procedure}(#{bind_placeholders(params)})"
+          connection.exec_update(statement, procedure, params)
+        end
+
+        def query_function(function, params, columns = ['*'])
```
lvonk, line 12: Idem for `function`

lvonk: Named arguments are preferred when they are optional: `columns: ['*']`

spec/lib/sequent/generator/project_spec.rb
```diff
@@ -58,7 +58,9 @@
     expect(File.read('blog-with_special-symbols/Rakefile')).to include("require './blog_with_special_symbols'")
   end
 
-  it 'has working example with specs' do
+  # Ignore for now until we find a way to correctly test the template project
```
lvonk, line 61: I think this is fixed by @werdlerk in master

lvonk (commented):
Some comments.

lvonk:
When I follow the instructions from `sequent migrate` I get the following error:

```
blog_development=# \i db/sequent_8_migration.sql
db/sequent_8_migration.sql: No such file or directory
```

This file is in the gem directory, not the project. We need a way to retrieve it, or point to the right directory when running.

lib/sequent/core/event_store.rb
erikrozendaal: The ` ActiveRecord.default_timezone` can only have the values `:local` or `:utc`, AFAIK. Here is some description of how Rails/ActiveRecord is configured for time zones: https://stackoverflow.com/a/32229086

spec/lib/sequent/generator/project_spec.rb
erikrozendaal: It breaks on my setup, some confusion with nested bundlers/`Gemfile`s, etc. I've not been able to fix this.

Also, the setup on master tests against the latest version published on rubygems, not the version that you are currently developing.

erikrozendaal:
> When I follow the instructions from `sequent migrate` I get the following error:
> 
> ```
> blog_development=# \i db/sequent_8_migration.sql
> db/sequent_8_migration.sql: No such file or directory
> ```
> 
> This file is in the gem directory, not the project. We need a way to retrieve it, or point to the right directory when running.

The database files are now copied when running the `sequent migrate` command.

erikrozendaal:
> 🔥 ❤️
> 
> As discussed:
> 
>     * We should have two versions of the upgrade script for upgrading to sequent 8: one without partitioning and one with partitioning

For now the script just generates a default partition by default, this is easier than introducing two separate scripts.
 
>     * This script should guide the user through the process

Done (`sequent migrate`)


>     * To avoid duplication we should split `sequent_schema.sql` into tables and indexes so it can be used for both initial starters as upgraders.

Done.

> Later:
> 
>     * We should have a script that helps user starting in a fresh sequent 8 and later want to introduce partitioning of the event store.

No longer necessary as the tables are always partitioned (even when there is just a single default partition when the scripts are not adjusted).

PR #435, opened 2024-12-19
Avoid exhausting type ids due to transaction rollbacks

erikrozendaal:
Type ids (for aggregate, command, and event types) use the `SMALLINT` data type to optimize storage usage. However, when a sequence is used to generated ids the ids can be dropped on transaction rollback. This can cause the ids to be quickly exhausted when something is misconfigured.

This change uses the highest id + 1 for the next id. This is slow and requires table locking, but since new types are rarely added this should not impact normal performance.

- [x] Add changelog and migration instructions


db/sequent_schema_tables.sql
```diff
@@ -1,6 +1,6 @@
-CREATE TABLE command_types (id SMALLINT GENERATED ALWAYS AS IDENTITY PRIMARY KEY, type text UNIQUE NOT NULL);
-CREATE TABLE aggregate_types (id SMALLINT GENERATED ALWAYS AS IDENTITY PRIMARY KEY, type text UNIQUE NOT NULL);
-CREATE TABLE event_types (id SMALLINT GENERATED ALWAYS AS IDENTITY PRIMARY KEY, type text UNIQUE NOT NULL);
+CREATE TABLE command_types (id SMALLINT PRIMARY KEY, type text UNIQUE NOT NULL);
```
lvonk, line 1: Can we have something like `migrate_metadata_tables` ? Or use AR migrations for this.

erikrozendaal: That would be nice at some point, for now I added a quick fix to the new pgpsql file that drops these IDENTITY declarations if they exist.

PR #437, opened 2024-12-24
Enforce user defined unique keys for aggregates

erikrozendaal:
TODO:

- [x] DSL for defining unique keys
- [x] More tests
- [x] Documentation
- [x] Changelog
- [x] Database migration


lib/sequent/core/aggregate_root.rb
```diff
@@ -98,12 +98,37 @@ def to_s
...
+      #
+      # An `AggregateKeyNotUniqueError` is raised if a unique
+      # constrained is violated when committing the events to the
+      # database.
+      #
+      # Sample return value:
+      #
+      # ```
+      # {
+      #   user_email: { email: 'bob@example.com' }
```
lvonk, line 118: I prefer something more descriptive like:

```
class User < AggregateRoot
  unique_constraint :email, scope: :user_email
  unique_constraint :first_name, :last_name, scope: :user_name
end
```

This will add two unique constraints.

The `email`, `first_name` and `last_name` are then methods/attr_readers on the aggregate that will called when committing the events to generate the internal structure.

lvonk: The `key` can also be an array of values instead of a hash.

lib/sequent/core/event_store.rb
```diff
@@ -250,8 +253,12 @@ def store_events(command, streams_with_events = [])
             Sequent::Core::Oj.dump(events),
           ],
         )
-      rescue ActiveRecord::RecordNotUnique
-        raise OptimisticLockingError
+      rescue ActiveRecord::RecordNotUnique => e
+        if e.message =~ /duplicate aggregate key value/
+          raise AggregateKeyNotUniqueError
```
lvonk, line 258: Can we add here what value is not unique?

lvonk (commented):
This is a really nice addition. I do prefer a more descriptive style (like ActiveRecord does) for the implementation. This also gives us more flexibility whenever we decide to for instance implement storing these keys differently.

lib/sequent/core/aggregate_root.rb
lvonk: Should order matter? In other words are:

`unique_constraint :first_name, :last_name, scope: :user_name` and `unique_constraint :last_name, :first_name, scope: :user_name` the same?

erikrozendaal: They are the same (we use a JSONB column to store the key and PostgreSQL normalizes objects by field name, etc). Arrays would be different though...

lib/sequent/core/event_store.rb
erikrozendaal: Not easily. We'd have to insert the unique keys one by one to and then somehow include it in the SQL exception message and parse it in the Ruby code. I think for 99% of the cases it will just be an aggregate with a single constraint (e.g. user with an email login) so the unique violation is mostly clear.

Of course, in more complex cases it might be useful for debugging. We can at least include the SQL error message in the exception message I guess, or a caused by?

lvonk: Ruby will automatically add the causing exception as `cause` of `AggregateKeyNotUniqueError` in this scenario.

erikrozendaal: Added at least the aggregate type and id to the error message.

PR #438, opened 2025-01-01
Support ActiveRecord/Rails 8.0

evsasse:
Would be awesome if we can also back-port this to a `v7.2.1` WDYT? @lvonk

### DIFF
--- a/lib/sequent/support/database.rb
+++ b/lib/sequent/support/database.rb
@@ -49,9 +49,6 @@ module Sequent
           db_config = db_config.deep_merge(
             Sequent.configuration.primary_database_key => db_config_overrides,
           ).stringify_keys
-          if ActiveRecord::VERSION::MAJOR == 7 && ActiveRecord::VERSION::MINOR < 1
-            ActiveRecord.legacy_connection_handling = false
-          end
           ActiveRecord::Base.configurations = db_config.stringify_keys
           ActiveRecord::Base.connects_to database: {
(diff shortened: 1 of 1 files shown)


lib/sequent/support/database.rb
```diff
@@ -49,7 +49,7 @@ def self.establish_connection(db_config, db_config_overrides = {})
           db_config = db_config.deep_merge(
             Sequent.configuration.primary_database_key => db_config_overrides,
           ).stringify_keys
-          if ActiveRecord::VERSION::MAJOR >= 7 && ActiveRecord::VERSION::MINOR < 1
+          if ActiveRecord::VERSION::MAJOR == 7 && ActiveRecord::VERSION::MINOR < 1
```
evsasse, line 52: This is the main change I needed to make it work here for now... Still exploring it a bit deeper against our own app.

erikrozendaal: We no longer support ActiveRecord before 7.1 so I think this `if` can be completely removed?

erikrozendaal (commented):
Thanks for your PR. Looks good to me.

lib/sequent/support/database.rb
evsasse: Done ✅

erikrozendaal:
@evsasse thanks! I've also created a [v7.x](https://github.com/zilverline/sequent/tree/v7.x) branch so you can make a PR to support ActiveRecord 8 if you wish. In this case the `if` around `legacy_connection_handling` will still be required.

PR #446, opened 2025-02-21
Register types rake task

erikrozendaal:
Adds a new Rake task `sequent:register:types` to insert aggregate root, command, and event types into the `*_types` tables. Useful to run after deploying new code so all new types are registered in advance.

### DIFF
--- a/lib/sequent/rake/migration_tasks.rb
+++ b/lib/sequent/rake/migration_tasks.rb
@@ -20,38 +20,8 @@ module Sequent
         Sequent::Support::Database.read_database_config(@env)
       end
 
-      def connection
-        ensure_sequent_env_set!
-        Sequent::Support::Database.establish_connection(db_config)
-      end
-
       def register_tasks!
         namespace :sequent do
-          namespace :install do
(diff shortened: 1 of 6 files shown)


lvonk (commented):
I think it is nice to have this also in a single class so it can be used in initializers. Rails apps can then check if preloading (a prerequisite for using Sequent) turned on:

https://sequent.io/docs/rails-sequent.html#eager-loading

lib/sequent/rake/migration_tasks.rb
```diff
@@ -36,6 +36,25 @@ def register_tasks!
...
+
+              NOTE make sure to load all Ruby classes before running this task!
+            EOS
+            task types: %i[sequent:init] do
+              ensure_sequent_env_set!
+
+              connection
+
+              Sequent.configuration.event_store.register_types(
+                aggregate_root_classes: all_subclasses(Sequent::Core::AggregateRoot),
```
lvonk, line 51: `Sequent::Core::AggregateRoot.descendants`

PR #454, opened 2025-03-21
Allow you to run the migration to Sequent 8 even if you have not updated `aggregate_id` to be UUID yet.

evsasse:
The columns on the tables expect a UUID, but if you have not updated your old tables to use UUID for `aggregate_id` you would hit errors. (`... is of type uuid but expression is of type character varying`).

AFAIK there is no penalty/problem in case you already are using the UUID, the casting would be a no-op.


erikrozendaal:
LGTM, thanks @evsasse !

lvonk (approved):

PR #457, opened 2025-04-07
Add read-only mode option during offline migration

mvandiepen:
Adds read-only mode and fails with a `ReadOnlyModeEnabled` error given a newer running/done offline migration.

### DIFF
--- a/lib/sequent/configuration.rb
+++ b/lib/sequent/configuration.rb
@@ -65,7 +65,8 @@ module Sequent
                   :primary_database_role,
                   :primary_database_key,
                   :time_precision,
-                  :enable_autoregistration
+                  :enable_autoregistration,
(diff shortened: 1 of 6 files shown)


lib/sequent/configuration.rb
```diff
@@ -127,6 +128,7 @@ def initialize
       self.time_precision = DEFAULT_TIME_PRECISION
 
       self.enable_autoregistration = false
+      self.enable_offline_migration_read_only_mode = false
```
mvandiepen, line 131: I think this might actually be good default behaviour, since the alternative is that the application is not running or the projections would have missing data.

mvandiepen: Also, to be documented on release.

PR #459, opened 2025-04-08
Register migrating resources on migration table

mvandiepen:
Register the resources affected in the migrations table. Also add some convenience scopes on versions.

### DIFF
--- a/lib/sequent/migrations/versions.rb
+++ b/lib/sequent/migrations/versions.rb
@@ -10,7 +10,9 @@ module Sequent
 
       def self.migration_sql
         <<~SQL.chomp
-          CREATE TABLE IF NOT EXISTS #{table_name} (version integer NOT NULL, CONSTRAINT version_pk PRIMARY KEY(version));
+          CREATE TABLE IF NOT EXISTS #{table_name} (version integer NOT NULL, target_projectors text[] DEFAULT '{}'::text[], target_records text[] DEFAULT '{}'::text[], CONSTRAINT version_pk PRIMARY KEY(version));
+          ALTER TABLE #{table_name} ADD COLUMN IF NOT EXISTS target_projectors text[] DEFAULT '{}'::text[];
+          ALTER TABLE #{table_name} ADD COLUMN IF NOT EXISTS target_records text[] DEFAULT '{}'::text[];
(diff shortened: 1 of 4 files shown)


lib/sequent/migrations/versions.rb
```diff
@@ -10,7 +10,9 @@ class Versions < Sequent::ApplicationRecord
 
       def self.migration_sql
         <<~SQL.chomp
-          CREATE TABLE IF NOT EXISTS #{table_name} (version integer NOT NULL, CONSTRAINT version_pk PRIMARY KEY(version));
+          CREATE TABLE IF NOT EXISTS #{table_name} (version integer NOT NULL, target_projectors text[] DEFAULT '{}'::text[], target_records text[] DEFAULT '{}'::text[], CONSTRAINT version_pk PRIMARY KEY(version));
+          ALTER TABLE #{table_name} ADD COLUMN IF NOT EXISTS target_projectors text[] DEFAULT '{}'::text[];
+          ALTER TABLE #{table_name} ADD COLUMN IF NOT EXISTS target_records text[] DEFAULT '{}'::text[];
```
erikrozendaal, line 15: It is now possible to use regular active record migrations in the `db/migrate` directory. The SQL here should actually be removed.

lib/sequent/migrations/versions.rb
```diff
@@ -64,7 +69,11 @@ def self.start_offline!(new_version)
           fail MigrationDone if current_migration.status.nil?
           fail ConcurrentMigration if current_migration.status != MIGRATE_ONLINE_FINISHED
 
-          current_migration.update(status: MIGRATE_OFFLINE_RUNNING)
+          current_migration.update(
+            status: MIGRATE_OFFLINE_RUNNING,
+            target_projectors:,
+            target_records:,
```
erikrozendaal, line 75: Maybe sort the arrays before updating, makes it slightly easier to read/debug the migration table later.

lib/sequent/migrations/versions.rb
mvandiepen: As discussed. We'll leave this as is for now since the table for `Sequent::Migrations::Versions` lives on the view schema.

PR #464, opened 2025-04-29
Fix view schema migration containing STI projections

lvonk:
Ensure all classes in a STI tree have the correct table_name
and column_information during the migration process.
Although `table_name = 'new_table_name'` internally also resets
all column information we are not sure this also happens
in all supported AR version. So to be sure we call `reset_column_information`
anyway.


mvandiepen (approved):

anya-nutt (approved):
Nice find!

erikrozendaal (approved):

PR #465, opened 2025-04-30
Activerecord transaction callbacks

erikrozendaal:
The old implementation was buggy (it would call all `after_commit`
callbacks when an inner transaction completed, before the actual
database commit happened), so replace it with the working and tested
ActiveRecord callbacks. Add support for `after_rollback` as well.


lvonk (commented):
Nice, this is a breaking change though as we drop support for 7.1. So we should release 9.

PR #468, opened 2025-05-08
Aggregate snapshot versioning

erikrozendaal:
To support changing the internals of an aggregate root without having
to throw away the old snapshots the snapshot version can now be
specified. This is needed to be able to support zero-downtime
deployments, where both old and new code run concurrently using the
same event store, but each may require different snapshots.

Example usage:

```
enable_snapshots version: 2
```

The default version is 1.

A hash mapping aggregate types to the supported snapshot version is
passed to every database call that needs it so the snapshot can be
correctly selected. The default implementation queries all descendants
of `AggregateRoot` for the supported snapshot version. This can be
changed using the `aggregate_snapshot_versions` configuration setting.

### DIFF
--- a/spec/lib/sequent/core/event_store_spec.rb
+++ b/spec/lib/sequent/core/event_store_spec.rb
@@ -10,6 +10,7 @@ describe Sequent::Core::EventStore do
   end
 
   class MyAggregate < Sequent::Core::AggregateRoot
+    enable_snapshots version: 42
   end
 
   let(:event_store) { Sequent.configuration.event_store }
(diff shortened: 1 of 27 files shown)


spec/lib/sequent/core/event_store_spec.rb
```diff
@@ -668,7 +680,7 @@ def stop
         streamed_events = event_store.load_events_for_aggregates([aggregate_id_1])
         expect(streamed_events).to have(1).items
         expect(streamed_events[0]).to have(2).items
-        expect(streamed_events[0][1]).to have(2).items
+        expect(pp(streamed_events[0][1])).to have(2).items
```
lvonk, line 683: `pp`

PR #469, opened 2025-05-12
Specify aggregate and event database insertion order

erikrozendaal:
When inserting multiple rows it is important that all inserts happen in a consistent order, otherwise multiple concurrent transactions can deadlock.

Also ensure newly inserted events have a correct sequence number so that a primary key constraint violation is raised when multiple concurrent transactions update the same aggregate, which will result in a Sequent optimistic locking error.

### DIFF
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -535,11 +535,13 @@ BEGIN
     _aggregate_id = _aggregate->>'aggregate_id';
     SELECT events_partition_key INTO STRICT _events_partition_key FROM aggregates WHERE aggregate_id = _aggregate_id;
 
-    SELECT MAX(sequence_number)
+    SELECT sequence_number
       INTO _last_sequence_number
       FROM events
      WHERE partition_key = _events_partition_key
-       AND aggregate_id = _aggregate_id;
+       AND aggregate_id = _aggregate_id
+     ORDER BY 1 DESC
+     LIMIT 1;
 
     SELECT MIN(event->'event_json'->>'sequence_number')
       INTO _next_sequence_number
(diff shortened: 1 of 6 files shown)


db/structure.sql
```diff
@@ -541,7 +562,8 @@ BEGIN
            _command_id,
            (SELECT id FROM event_types WHERE type = event->>'event_type'),
            (event->'event_json') - '{aggregate_id,created_at,event_type,sequence_number}'::text[]
-      FROM jsonb_array_elements(_events) AS event;
+      FROM jsonb_array_elements(_events) AS event
+     ORDER BY 1, 2;
```
lvonk, line 566: `1,2,3`

db/structure.sql
```diff
@@ -533,6 +535,25 @@ BEGIN
...
+      INTO _next_sequence_number
+      FROM jsonb_array_elements(_events) AS event;
+
+    -- Check sequence number of first new event to ensure optimistic locking works correctly
+    -- (otherwise two concurrent transactions could insert events with different first/next
+    -- sequence number and no constraint violation would be raised).
+    IF _last_sequence_number IS NULL AND _next_sequence_number <> 1 THEN
+      RAISE EXCEPTION 'sequence_number of first event must be 1';
+    ELSIF _last_sequence_number IS NOT NULL AND _next_sequence_number <= _last_sequence_number + 1 THEN
+      RAISE EXCEPTION 'sequence_number must be consecutive';
```
lvonk, line 554: It would be handy if we would the sequence numbers in the error message

lvonk (approved):
lgtm

PR #472, opened 2025-06-10
Track projectors state (replaying/activating/active) in the database

erikrozendaal:
The event publisher and projector check the current state when processing events to:

- avoid publishing any events if there is an unknown projector active (to avoid running old code on upgraded systems)
- prevent processing events for a projector where a newer version, so there is no need to shut down the old code before starting up the new code. A load balancer should redirect requests the new code as soon as possible after activating the newer projector since the old code will start failing with `NewerProjectorIsActiveError` errors.

This PR is another step towards allowing rolling upgrades, where version N and version N+1 code is running concurrently during an upgrade, without affecting any users.

TODO
- [x] document need for `activate_current_configuration!`
- [x] changelog

### DIFF
--- a/lib/sequent/generator/template_project/spec/spec_helper.rb
+++ b/lib/sequent/generator/template_project/spec/spec_helper.rb
@@ -40,7 +40,7 @@ RSpec.configure do |config|
 
   config.around do |example|
     Sequent.configuration.aggregate_repository.clear
-    DatabaseCleaner.strategy = [:truncation, {except: %w[projector_states]}]
+    DatabaseCleaner.strategy = :transaction
     DatabaseCleaner.cleaning do
       example.run
     ensure


lib/sequent/core/projector.rb
```diff
@@ -193,24 +198,21 @@ def ensure_valid!
...
 
-    #
-    # Utility class containing all subclasses of Projector.
-    #
-    class Projectors
-      class << self
-        def projectors
-          Sequent::Projector.descendants
-        end
+      def is_active?
```
lvonk, line 202: More rubish.

```suggestion
      def active?
```

lib/sequent/core/event_store.rb
```diff
@@ -173,7 +173,9 @@ def events_exists?(aggregate_id)
       def replay_events
         warn '[DEPRECATION] `replay_events` is deprecated in favor of `replay_events_from_cursor`'
```
lvonk, line 174: Perhaps remove this method as this has been deprecated for long?

lib/sequent/core/event_publisher.rb
```diff
@@ -30,6 +35,15 @@ def message
       def publish_events(events)
         return if configuration.disable_event_handlers
 
+        ensure_no_unknown_active_projectors!
+
+        events.each { |event| events_queue.push(event) }
+        process_events
+      end
+
+      def replay_events(events)
```
lvonk, line 44: Maybe this should be in an `EventReplayer` or so?

lib/sequent/core/event_store.rb
```diff
@@ -188,14 +190,18 @@ def replay_events_from_cursor(get_events:, block_size: 2000,
...
-          ids_replayed << record['aggregate_id']
-          if progress % block_size == 0
-            on_progress[progress, false, ids_replayed]
-            ids_replayed.clear
+        cursor.each_row_batch(block_size:).each do |records|
+          events = records.map(&method(:deserialize_event))
+          Sequent.configuration.transaction_provider.transactional do
+            Sequent.configuration.event_publisher.replay_events(events)
+          end
+          records.each do |record|
```
lvonk, line 198: This is not really true anymore as this is logging after the fact. We should probably rethink logging

lib/sequent/sequent.rb
```diff
@@ -37,17 +37,27 @@ def self.configuration
     Configuration.instance
   end
 
+  def self.activate_current_configuration!
+    configuration.event_store.register_types!
+    migration_class&.activate_current_configuration!
```
lvonk, line 42: `migrations_class`

PR #474, opened 2025-07-01
Check projector active state using a specialized event publisher

erikrozendaal:
This simplifies the projector logic as all projector state tracking is now only checked by the `ActiveProjectorsEventPublisher`. Replay now uses the normal `EventPublisher` so the logic to determine if an event is allowed to be processed by a projector can be simplified.

- [ ] Should we remove the `enable_projector_states` flag and just require the user to configure the event publisher as `ActiveProjectorsEventPublisher`?


mvandiepen (approved):
LGTM ✅

PR #500, opened 2026-01-16
Early exit from a message handler should not skip subsequent handlers

erikrozendaal:
~~Raise an error to notify the developer to use `next` instead of
`return` to skip (part of) the message handler code.~~

Use `define_method` to make event handler blocks behave like 
normal Ruby methods, so using `return` just returns from the
message handler body and does not exit the event handling loop
of the message dispatcher.

In Ruby the `return` from a block returns from the enclosing
method. Since we capture blocks when registering handlers with `on`
and then execute them later using `instance_exec` this causes
unintuitive behaviour, as subsequent handlers are no longer executed
when a handler uses `return`.


lvonk (approved):
Nice catch!

PR #503, opened 2026-01-21
WIP: add stream name to event

Braitenberg:
so that we can change the class names while maintaining the internal naming of an event


lvonk:
What is the rationale behind `stream_name`? An event is not a stream (which is a series of events).

PR #526, opened 2026-06-10
New snapshot versions are marked as outdated on registration

erikrozendaal:
This ensures these aggregates will be snapshotted again so the new version is stored.

Also allow storing a snapshot for the same aggregate, version, and sequence number by simply overwriting the existing snapshot.

### DIFF
--- a/db/migrate/20260610094300_sequent_fix_snapshot_version_upgrades.rb
+++ b/db/migrate/20260610094300_sequent_fix_snapshot_version_upgrades.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+class SequentFixSnapshotVersionUpgrades < ActiveRecord::Migration[7.2]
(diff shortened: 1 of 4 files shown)


db/migrate/sequent/store_snapshots_v02.sql
```diff
@@ -27,7 +27,11 @@ BEGIN
       (_snapshot->>'created_at')::timestamptz,
       _snapshot->>'snapshot_type',
       _snapshot->'snapshot_json'
-    );
+    )
```
lvonk: Why change v2?

lvonk (approved):

db/migrate/sequent/store_snapshots_v02.sql
erikrozendaal: Accident

PR #532, opened 2026-07-12
Update building-a-web-application.md

alexandru-calinoiu:
One needs to connect to the database, the actual error received is `ActiveRecord::ConnectionNotDefined: No database connection defined`


alexandru-calinoiu:
Dropping this because the docs actually refer to the master version that actually works with the current tutorials.

PR #533, opened 2026-07-24
Observers are invoked immediately when event is applied to aggregate

erikrozendaal:
This allows decoupling domain code and/or replacing simple workflows.


lib/sequent/core/observer.rb
```diff
@@ -0,0 +1,52 @@
...
+    # Utility class containing all subclasses of Observer.
+    #
+    class Observers
+      class << self
+        def observers
+          Sequent::Observers.descendants
+        end
+
+        def all
+          workflows
```
lvonk: `observers`

lib/sequent/core/observer.rb
```diff
@@ -0,0 +1,52 @@
...
+      end
+    end
+
+    #
+    # Utility class containing all subclasses of Observer.
+    #
+    class Observers
+      class << self
+        def observers
+          Sequent::Observers.descendants
```
lvonk: `Sequent::Observer.descendants`

lvonk (commented):
Looks good. Perhaps `AggregateObserver` like we discussed, to make the scope clear and when it is invoked.

